PR #1, opened 2015-10-28
Initial commit

mathiasquintero:
Starting with a basic frame. Nothing really implemented but already defining the view of one of the cards and the tabs in the bottom.

screenshot-initial-commit


mathiasquintero:
I personally don't like to approve my own pull requests. So I will wait a day for comments and then if nothing is wrong merge it.

PR #37, opened 2016-12-05
RoomFinder: Use the map with the smallest scale as the default map

### LINKED ISSUE #33: RoomFinder: Use map with smaller scale as default
Currently the RoomFinder automatically shows the Munich map when searching for a room. I'd assume most users rather want to see a smaller map, so they find the room inside a building - and not the building inside Munich.

So I guess it makes sense to choose a smaller scaled map by default.

mammuth:
Furthermore the list of available maps is sorted by their scale

Closes #33

### DIFF
--- a/TUM Campus App/Map.swift
+++ b/TUM Campus App/Map.swift
@@ -11,10 +11,12 @@ class Map: ImageDownloader, DataElement {
     let roomID: String
     let mapID: String
     let description: String
-    init(roomID: String, mapID: String, description: String) {
+    let scale: Int
+    init(roomID: String, mapID: String, description: String, scale: Int) {
         self.roomID = roomID
         self.mapID = mapID
         self.description = description
+        self.scale = scale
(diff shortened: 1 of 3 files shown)


mathiasquintero:
Is taking the smallest the best solution? I'm pretty sure there is a default map api endpoint

mammuth:
Probably it's not the best solution for everyone. But I'd say for most users it is. 

From the smallest scale students can recognize the building (Arcisstraße, FMI, MW, ...) and mostly also know where this building is roughly located. 

In my case, I don't use the room finder often, but when I do, I want to find the room within a building.

kordianbruck:
@mammuth yes there is an API for the default map. "*/room/defaultMap/*"

https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/tumonline/TUMRoomFinderRequest.java

mammuth:
@mathiasquintero you're still self-assigned, are you doing something on this topic right now?

mathiasquintero:
Ah! I forgot! Do you I think we can merge this and I can will implement the default map

PR #42, opened 2016-12-16
Show upcoming news on news card and news table view

### LINKED ISSUE #40: News: Card and list is showing most future news first
The news-card is currently showing the event that's at position one of the api response. Currently that's a TU movie anytime in 2017. 
The card should probably show the news which is the **next upcoming one**?

Same for the list of news - in Android this is solved by scrolling to the position of the closest news.

mammuth:
Before this PR, the newscard always displayed the newest story, now it displays the next upcoming story.
Also the collection view of all news (More>News) scrolls to the next upcoming story. This is inspired by the behaviour of the android app.

Closes #40

### DIFF
--- a/TUM Campus App/NewsManager.swift
+++ b/TUM Campus App/NewsManager.swift
@@ -56,12 +56,14 @@ class NewsManager: Manager {
     }
     
     func handleNews(_ handler: ([DataElement]) -> ()) {
-        let items = NewsManager.news.sorted { (a, b) in
-            return a.date.compare(b.date as Date) == ComparisonResult.orderedDescending
-        }
         if single {
-            handler([getNextUpcomingNews()])
+            if let story = getNextUpcomingNews() {
+                handler([story])
+            }
         } else {
+            let items = NewsManager.news.sorted { (a, b) in
+                return a.date.compare(b.date as Date) == ComparisonResult.orderedDescending
+            }
             var returnableArray = [DataElement]()
             for item in items {
                 returnableArray.append(item)
@@ -70,14 +72,12 @@ class NewsManager: Manager {
         }
     }
     
-    func getNextUpcomingNews() -> News {
+    func getNextUpcomingNews() -> News? {
         let now = Date()
-        for (index, item) in NewsManager.news.enumerated() {
-            if item.date < now  {
-                return NewsManager.news[index-1]
-            }
+        if let firstStory = NewsManager.news.filter({ $0.date > now }).last ?? NewsManager.news.first {
+            return firstStory
         }
-        return NewsManager.news[0] // There are only past news
+        return nil
     }
     
     func getURL() -> String {
--- a/TUM Campus App/NewsManager.swift
+++ b/TUM Campus App/NewsManager.swift
@@ -60,9 +60,7 @@ class NewsManager: Manager {
             return a.date.compare(b.date as Date) == ComparisonResult.orderedDescending
         }
         if single {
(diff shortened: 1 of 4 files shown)


TUM Campus App/NewsManager.swift
```diff
@@ -60,9 +60,7 @@ class NewsManager: Manager {
             return a.date.compare(b.date as Date) == ComparisonResult.orderedDescending
         }
         if single {
-            if let firsStory = items.first {
```
mathiasquintero: This could crash if the news array is empty... My suggestion:

```Swift
if let firstStory = items.filter({ $0.date > now }).last ?? items.first {
    handler([firstStory])
}
```

Or make the result of getNextUpcomingNews an optional. But I would prefer to never have anything with indexes in the code. Makes it safer

TUM Campus App/NewsTableViewController.swift
```diff
@@ -37,6 +37,7 @@ extension NewsTableViewController {
         tableView.estimatedRowHeight = tableView.rowHeight
         tableView.rowHeight = UITableViewAutomaticDimension
         tableView.tableFooterView = UIView(frame: CGRect.zero)
+        tableView.scrollToRow(at: IndexPath(row: getNextUpcomingNewsRow(), section: 0), at: UITableViewScrollPosition.top, animated: false)
```
mathiasquintero: You don't like it animated?

mammuth: Well, I thought the animation was to fast when there are a a couple of news item before the current one. Since scrollToRow is called in viewDidLoad (and not viewDidAppear), the user wont see the list jumping, but directly sees the wished story on top of the screen.
Personally I'd keep it like this.

mammuth:
> This could crash if the news array is empty...
You're right, thanks for the heads up. I changed _getNextUpcomingNews()_ to be optional and tested that nothing crashs when the news array is empty. 

Filtering is definitely more elegant 👍

TUM Campus App/NewsManager.swift
```diff
@@ -70,14 +72,12 @@ class NewsManager: Manager {
...
     }
     
-    func getNextUpcomingNews() -> News {
+    func getNextUpcomingNews() -> News? {
         let now = Date()
-        for (index, item) in NewsManager.news.enumerated() {
-            if item.date < now  {
-                return NewsManager.news[index-1]
-            }
+        if let firstStory = NewsManager.news.filter({ $0.date > now }).last ?? NewsManager.news.first {
```
mathiasquintero: you don't need this if let

TUM Campus App/NewsManager.swift
```diff
@@ -72,6 +72,14 @@ class NewsManager: Manager {
         }
     }
     
+    func getNextUpcomingNews() -> News? {
+        let now = Date()
+        if let firstStory = NewsManager.news.filter({ $0.date > now }).last ?? NewsManager.news.first {
```
mathiasquintero: Other suggestion. Take the news array as an optional parameter defaulting to NewsManager.news.

Because I feel like we have this might need this for filtered subsets from: https://github.com/TCA-Team/iOS/issues/43

```Swift
func getNextUpcomingNews(in news: [News] = NewsManager.news) { ... }
```

mathiasquintero: And also, you don't need the if let here ;)

mammuth:
Added your latest suggestion 👍

PR #44, opened 2016-12-16
Refactoring a bit

mathiasquintero:
This code makes the data persistence of the log in a bit more straight forward by no longer storing two strings but rather a dictionary.

This also adds a dependency that I myself maintain: 'Sweeft' (Super proud of it)

I was also having some issues because of the spacing in the target name. Xcode 8.2 wasn't noticing the spaces and treated them as separate arguments on sh scripts.


mammuth:
On this revision, no news images are shown in my simulator.

bildschirmfoto 2016-12-17 um 14 47 21

mathiasquintero:
That's a strange bug. I'll check it out

mammuth:
If it's something that doesn't occur for you, drop me a line then I try to investigate on my side aswell.

mathiasquintero:
It is happening. I get no response from the server.

Try it: http://www.tu-film.de/img/film/poster/.thumb.How%20to%20Be%20Single.jpg

mammuth:
Seems like their machine is down.

bildschirmfoto 2016-12-17 um 15 52 15

PR #46, opened 2016-12-17
Feature/allow saving cards and order

mathiasquintero:
DO  NOT MERGE without merging https://github.com/TCA-Team/iOS/pull/44 since this branch is based on that one.

You know... Git Magic ;)


mammuth:
"Delete" should probably named "disable" so the user understands that this card is going to be moved to the section of disabled ones below.
bildschirmfoto 2016-12-17 um 16 10 40
Do we even need this confirmation? The "-" icon is clear and if someone disables a card on accident, it's as easy to restore as one single click. So in my eyes the confirmation is rather annoying than helpful.
Can the confirmation easily skipped or would that be unnecessary work?

And as stated in the review: If I disable all cards, the app crash when going back to the card view and refresh them. Nothing spectacular, since hopefully no one will ever come to this idea.

mathiasquintero:
Added this beautiful card for when there are no cards:

simulator screen shot dec 17 2016 4 44 18 pm

mammuth:
Cool! 🎉

mathiasquintero:
@mammuth it's actually easier if you drag them down. I added it as an extra option and because it looked weird without the (-) button.

PR #47, opened 2016-12-17
Add room finder to more view

### LINKED ISSUE #34: Add RoomFinder to "More" View
Currently the only way to search for rooms is by using the search. A new user however doesn't know this feature exists.

There should be a reference in the more-view. I guess it might just open the search bar?

mammuth:
Before there was no hint to the user, that the room finder is implemented. Now it's referenced in the more-view.
SearchViewController is restricted to the RoomSearch when opened via the more view.

@mathiasquintero where are the current icons from? We probably should add a dedicated icon for the roomfinder?

Closes #34 

bildschirmfoto 2016-12-17 um 14 33 13


TUM Campus App/MoreTableViewController.swift
```diff
@@ -76,7 +81,7 @@ extension MoreTableViewController {
             }
         case 5:
             UserDefaults.standard.removeObject(forKey: LoginDefaultsKeys.Token.rawValue)
-            let storyboard = UIStoryboard(name: "Main", bundle: nil)
+            let storyboard = UIStoryboard(name: "Setup", bundle: nil)
```
mathiasquintero: Wasn't this done before?

TUM Campus App/TumDataManager.swift
```diff
@@ -77,9 +77,10 @@ class TumDataManager {
         }
     }
     
-    func search(_ receiver: TumDataReceiver, query: String) {
+    func search(_ receiver: TumDataReceiver, query: String, searchIn managersToSearchIn:[TumDataItems]? = nil) {
```
mathiasquintero: Why not = searchManagers? That way you don't have to check for an optional

TUM Campus App/MoreTableViewController.swift
mammuth: Not in the MoreTableViewController. But somewhere else, yes.

TUM Campus App/TumDataManager.swift
mammuth: Wanted to do so, but Xcode gave an error that it's not possible to use instance variables as default parameters.

mathiasquintero: And if we make it static? It makes sense for it to be static anyway

mathiasquintero: Let's leave it be for now

PR #48, opened 2016-12-17
Add explanation on how to authorize the app on TUM Online

### LINKED ISSUE #45: Better explanation of the authorization step
In Android there is a image showcasing what to do in TUMOnline, we should probably add similar explanations before releasing it.

bildschirmfoto 2016-12-17 um 11 14 28

mammuth:
It's not as detailed as in Android, but most people will understand what to do and we don't need to bother with this for the MVP.

bildschirmfoto 2016-12-17 um 15 23 47

Closes #45


mathiasquintero:
Makes sense! Good thinking. Let's merge this and consider adding a screenshot of TUMOnline later. Because I never read the instructions ;)

PR #50, opened 2016-12-19
Plans feature

### LINKED ISSUE #26: Plans
- MVV Tram
- MVV complete
- Campus Garching
- Campus Klinikum
- Campus Olympiapark
- Campus Olympiapark Hallenplan
- Campus Arcisstraße
- Campus Weihenstephan

TheZoker:
Closes #26


mammuth:
Cool, gonna check it out on the simulator later.

Looks like you're downloading the needed plans, but why are there binaries shipped with the app then? (https://github.com/TheZoker/iOS/blob/plans-rebase/TUM%20Campus%20App/Assets.xcassets/plan_campus_garching.imageset/campus_garching.png)

TheZoker:
I'm just doing the same as the android TUM Campus App:
https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/activities/PlansActivity.java#L33-L37

https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/activities/PlansActivity.java#L101-L111

kordianbruck:
The University Plans are shipped with the app. It is planed to serve them via an Endpoint over the API as well, but this is not yet implemented and would require some backend work.

@TheZoker please name your branches like so in the future: 'feature/<thename>' (Check this SO answer from 2008 http://stackoverflow.com/questions/273695/git-branch-naming-best-practices)

TUM Campus App/PlansTableViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+//  Copyright © 2016 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+import Alamofire
+
+class PlansTableViewController: UITableViewController {
+    
+    // MARK: - Table view data source
+    lazy var plans = Plans()
```
mathiasquintero: Should this really be evaluated every single time you load the view controller? Perhaps create a shared instance in Plans

TUM Campus App/PlansTableViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+            let nextScene = segue.destination as! PlanDetailsViewController
+            let plan = sender as! Plan
+            nextScene.planFileUrl = plan.fileUrl
+            nextScene.planType = plan.type
+        }
+    }
+    
+    // download the plans
+    private func downloadData() {
+        for i in 0..<plans.getPlans().count {
```
mathiasquintero: There's no need to go by index. Use foreach or for plan in plans.getPlans()

TUM Campus App/Plans.swift
```diff
@@ -0,0 +1,34 @@
...
+        plans.append(Plan(title: "Campus Klinikum", type: "image", fileUrl: "plan_campus_klinikum", url: "", address: "Ismaninger Straße 22, 81675 München", icon: "plan_campus_klinikum_icon"))
+        plans.append(Plan(title: "Campus Olympiapark", type: "image", fileUrl: "plan_campus_olympiapark", url: "", address: "Connollystraße 32, 80809 München", icon: "plan_campus_olympiapark_icon"))
+        plans.append(Plan(title: "Campus Olympiapark Hallenplan", type: "image", fileUrl: "plan_campus_olympiapark_hallenplan", url: "", address: "Connollystraße 32, 80809 München", icon: "plan_campus_olympiapark_hallenplan_icon"))
+        plans.append(Plan(title: "Stammgelände", type: "image", fileUrl: "plan_campus_stammgelaende", url: "", address: "Arcisstraße 21, 80333 München", icon: "plan_campus_stammgelaende_icon"))
+        plans.append(Plan(title: "Campus Weihenstephan", type: "image", fileUrl: "plan_campus_weihenstephan", url: "", address: "Liesel-Beckmann-Straße 6, 85354 Freising", icon: "plan_campus_weihenstephan_icon"))
+    }
+    
+    func getPlans() -> [Plan]{
```
mathiasquintero: It's not very swift to use setters and getters as functions. That's why set{ } and  get { } exist. Plus you can achieve this by having a ```private(set) var plans = [Plan]()``` which makes a getter internal but the setter private

TUM Campus App/PlanDetailsViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+//  Created by Florian Gareis on 16.12.16.
+//  Copyright © 2016 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+
+class PlanDetailsViewController: UIViewController, UIScrollViewDelegate {
+
+    @IBOutlet weak var scrollView: UIScrollView!
+    var imageView: UIImageView!
```
TheZoker: Yes I use it more than once: 
https://github.com/TCA-Team/iOS/pull/50/files#diff-eb09460ec53bb01303365b20ed9ba5b1R47

TUM Campus App/PlanDetailsViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+            let documentUrl = paths?.appendingPathComponent("plans/"+planFileUrl)
+            webView.loadRequest(NSURLRequest(url: documentUrl!) as URLRequest)
+            
+            scrollView.addSubview(webView)
+        }
+    }
+    
+    // allow image zooming
+    func viewForZooming(in scrollView: UIScrollView) -> UIView? {
+        return self.imageView
```
mathiasquintero: Since imageView is a UIImageView! won't it crash if we added a webview?

TUM Campus App/PlanDetailsViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+            self.scrollView.minimumZoomScale = 1.0
+            self.scrollView.maximumZoomScale = 5.0
+            scrollView.addSubview(imageView)
+        } else if (planType == "pdf") {
+            webView = UIWebView()
+            webView.scalesPageToFit = true
+            webView.frame = self.view.frame
+            
+            let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
+            let documentUrl = paths?.appendingPathComponent("plans/"+planFileUrl)
```
mathiasquintero: Please use proper spacing ```"plans/" + planFileUrl``` is much easier to read.

TUM Campus App/PlansTableViewCell.swift
```diff
@@ -0,0 +1,17 @@
...
+//  TUM Campus App
+//
+//  Created by Florian Gareis on 14.12.16.
+//  Copyright © 2016 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+
+class PlansTableViewCell: UITableViewCell {
+
```
mathiasquintero: Not really that important but there's an extra free line here. It's bugging the hell out of me

TUM Campus App/Plans.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Copyright © 2016 LS1 TUM. All rights reserved.
+//
+
+import Foundation
+
+class Plans {
+    private var plans = [Plan]()
+    
+    init(){
+        plans.append(Plan(title: "MVV Gesamtnetz", type: "pdf", fileUrl: "Schnellbahnnetz.pdf", url: "http://www.mvv-muenchen.de/fileadmin/media/Dateien/plaene/pdf/Netz_2017_Version_A4.pdf", address: "", icon: "plan_mvv_icon"))
```
mathiasquintero: Why do we sometimes have empty strings as values? If they can be empty perhaps they should be optionals and not be specified in the init

TUM Campus App/PlanDetailsViewController.swift
mathiasquintero: Is there any way for you to get rid of these extra properties? It doesn't seem very clean to keep a reference to these views

TUM Campus App/PlanDetailsViewController.swift
TheZoker: Works fine in my tests

TUM Campus App/PlanDownloader.swift
```diff
@@ -0,0 +1,32 @@
...
+        let fileURL = documentsURL!.appendingPathComponent("plans/"+name)
+        
+        let destination: DownloadRequest.DownloadFileDestination = { _, _ in
+            return (fileURL, [.removePreviousFile, .createIntermediateDirectories])
+        }
+        
+        let fileManager = FileManager.default
+        if !(fileManager.fileExists(atPath: fileURL.path)) {
+            Alamofire.download(url, to: destination).response { response in
+                print(response)
```
TheZoker: I'm not sure if I need a callback here. Maybe I should just ignore the callback?

TUM Campus App/PlansTableViewController.swift
TheZoker: Do you have any reference, how to built such a shared instance? I'm not really sure how to build them

TUM Campus App/Plans.swift
TheZoker: But how do I get the count value then? Like I do here: https://github.com/TCA-Team/iOS/pull/50/files#diff-5aacaf475b874319617ff1d168b5fc27R31

TUM Campus App/Plans.swift
TheZoker: It's just done like in the Android App:
https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/activities/PlansActivity.java#L105

But instead they use `R.string.empty_string`.

TUM Campus App/PlansTableViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    // load data into the rows
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let plan = plans.getPlan(forIndex: indexPath.row)
+        
+        let cell = tableView.dequeueReusableCell(withIdentifier: "planCell", for: indexPath) as! PlansTableViewCell
+
+        // Configure the cell...
+        cell.planTitleLabel.text = plan.title
```
TheZoker: I searched for tutorials on this, but every tutorial I found, configures the cell in the VC:

- https://www.ralfebert.de/tutorials/ios-swift-uitableviewcontroller/custom-cells/
- http://www.codingexplorer.com/getting-started-uitableview-swift/#crayon-5887c31500f88513863969
- http://shrikar.com/uitableview-and-uitableviewcell-customization-in-swift/

TUM Campus App/Plans.swift
mathiasquintero: You can do it straight from the variable. What private(set) means is that you can access the variable inside your module but only set it from inside the declaration. Which is exactly what you want. You would do .plans.count

TUM Campus App/PlanDownloader.swift
mathiasquintero: The question is shouldn't someone be notified when the download is done?

TUM Campus App/PlanDetailsViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+    var imageView: UIImageView!
+    var webView: UIWebView!
+    var planFileUrl: String!
+    var planType: String!
+    
+    // display image or pdf, depending on the file type
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        if (planType == "image") {
```
TheZoker: Fixed!

TUM Campus App/PlanDetailsViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+            self.scrollView.maximumZoomScale = 5.0
+            scrollView.addSubview(imageView)
+        } else if (planType == "pdf") {
+            webView = UIWebView()
+            webView.scalesPageToFit = true
+            webView.frame = self.view.frame
+            
+            let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
+            let documentUrl = paths?.appendingPathComponent("plans/"+planFileUrl)
+            webView.loadRequest(NSURLRequest(url: documentUrl!) as URLRequest)
```
TheZoker: Fixed!

TUM Campus App/PlansTableViewController.swift
TheZoker: Fixed!

TUM Campus App/PlanDetailsViewController.swift
TheZoker: Fixed!

TUM Campus App/PlansTableViewCell.swift
TheZoker: Fixed!

TUM Campus App/PlansTableViewController.swift
mathiasquintero: I know that a lot of tutorials do it this way. But this contributes to the horrible problem in iOS: Massive View Controller. The more modular your UI code is, the easier it will be to maintain it.

TUM Campus App/PlansTableViewController.swift
mathiasquintero: The common way is to have this in Plans: static lazy var shared = Plans()
It's a basic singleton. That way you don't have to load it every time. It's commonly done in FileManager, UIApplication and other Apple Framework Classes

TUM Campus App/Plans.swift
TheZoker: screenshot_2017 01 24_23h46m45s_022_

And [here](https://github.com/TCA-Team/iOS/pull/50/files#diff-5aacaf475b874319617ff1d168b5fc27R73)?
Is it like `for plan in .plans` ?

I'm very sorry for all the questions, but since I'm more a php and java developer, I'm not used to program the swift way yet.

TUM Campus App/PlansTableViewController.swift
TheZoker: Do you have a tutorial for that?

TUM Campus App/Plans.swift
mathiasquintero: I'll simplify. In Swift we never write setters and getters. Too repetitive. 
There are, computed, weak, lazy and of course the normal properties you are used to. So we manage access differently. One way is by adding the access modifier private(set) we say only code inside our class declaration can modify this variable. This is the equivalent to in Java only writing a getter and no setter.
What I mean is that you can get rid of getPlans() this way. Because everyone can read the array directly. So the count is in {PlansObject}.plans.count instead of {PlansObject}.getPlans().count

TUM Campus App/PlansTableViewController.swift
mathiasquintero: Not really since this is still a really controversial topic today. But you can see it in the rest of the app: https://github.com/TCA-Team/iOS/blob/master/TUM%20Campus%20App/TuitionCardTableViewCell.swift

As you can see there's an extra Object for the cell that is optimized for displaying it.
BTW: I know there's a setter here. I haven't updated some of the old code to the new style guidelines ;)

TUM Campus App/PlansTableViewController.swift
mathiasquintero: So my suggestion is:

In Plans:

```Swift
class Plans {

    static lazy var shared = Plans()

    // Rest of your code
}
```

And in your TableView instead of this plans variable you'd have:
```Swift
class PlansTableViewController: UITableViewController {

    var plans: [Plan] {
        return Plans.shared.plans
    }

}
```

And in your loop you just do:


```Swift
for plan in plans {
    if (plan.type == .pdf) {
        let downloader = PlanDownloader()
        downloader.downloadPlan(urlString: plan.url, withName: plan.fileUrl)
     }
}
```

Or if you're as lazy as me:

```Swift
plans
  .filter { $0.type == .pdf }
  .forEach { PlanDownloader().downloadPlan(urlString: $0.url, withName: $0.fileUrl) }
```

TheZoker: Thank you very much for your patience! Now I think I understand the getter and setter thing :)

PR #51, opened 2017-01-12
Update pods dependencies to latest

TheZoker:


mathiasquintero:
I don't see any changes in our code. I think there might be a problem with SwiftyJSON.JSON and Sweeft.JSON having the same name and being imported in the same file.

Does it build?

TheZoker:
Yea it build on my end without any issues

mathiasquintero:
BTW. You're welcome to help out on Sweeft ;) https://github.com/mathiasquintero/Sweeft

TheZoker:
I will have a look at it, but since I'm not a swift expert and still trying to learn it, I'm not sure, if I could be helpful here.

PR #58, opened 2017-03-06
Add tests for movie manager

mammuth:
Related #57

### DIFF
--- a/TUM Campus AppUnitTests/MovieTests.swift
+++ b/TUM Campus AppUnitTests/MovieTests.swift
@@ -20,6 +20,20 @@ class MovieTests: XCTestCase {
         super.tearDown()
     }
     
+    func testReceivedDataAreMovies() {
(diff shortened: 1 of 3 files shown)


TUM Campus AppUnitTests/MovieTests.swift
```diff
@@ -25,29 +20,36 @@ class MovieTests: XCTestCase {
...
-    
-    func receiveData(_ data: [DataElement]) {
-        for element in data {
-            if let movieElement = element as? Movie {
-                movies.append(movieElement)
+    func testOnlyUpcomingMovies() {
+        func receiveData(_ data: [DataElement], expectation: XCTestExpectation) {
+            let movies = data.flatMap { $0 as? Movie }
+            let dateNow = Date()
+            let upcomingMovies = movies.filter() { movie in
```
mathiasquintero: In the future. If the method only has a trailing closure omit the () just like in flatMap a few lines up

PR #75, opened 2017-04-11
Spinner fix

TG908:
Checks if the number of cards while refreshing and stops refreshing when no cards are present.


TG908:
Something is wrong with my build settings

mathiasquintero:
I don't really like a PR for a bugfix to update dependencies and change the project settings. This is gonna cause way too many conflicts for something that wasn't necessary for your fix. In the past, we've had PRs for updating pods and updating settings, etc completely separate. 

Because it's very hard to see the real changes here.
@mammuth what do you think? Should he cherry pick this into another branch? Or do we just allow this once?

TG908:
The original Master Branch didn't work when I pulled it and started my work on this project so I had to create a working master for me to even run the app on the simulator.

mathiasquintero:
We're currently considering putting pods on the .gitignore after these merges anyway. Because then we can update and install as much as we like and it won't appear on any PR aside from the .fixed file. As long as everyone always fetches the latest dependencies when working this kind of thing should be fine

PR #80, opened 2017-04-16
Add a “Not Yet Implemented” alert

mammuth:
When clicking on the following rows in the more view, a alert will be presented for the user:
- MVV
- Services
- Default Campus

image

This is meant as a quick and dirty fix so the user understand, why nothing happens. I prefer this solution over removing the items.

Related #70


mathiasquintero:
I like it!

Issue #64, opened 2017-03-27
Allow non TUM Students to use the app


kordianbruck:
We need to allow to use the App without the TUM ID asap.

Who has time to do this in the current week?

mammuth:
Maybe also add 1-2 additional cards with Infos that are available without a lrz account. 

How about a static card with links to resources for interested pupils about the studies at the TUM? This card can just be hidden when there is a logged in user.

mathiasquintero:
@mammuth I love the idea of tum.sexy. Where can I just get the links without the page?

kordianbruck:
Check the redirection/ folder in the repo. There is a routes.php file there containing only the redirections

mathiasquintero:
Yeah, but how can I best fetch them from the phone?

kordianbruck:
We could add an endpoint that would return the json? Would that work?

mathiasquintero:
That would be awesome

mammuth:
I'm off for work today, Kordian can you do the json endpoint?

kordianbruck:
Yup, will do in 45 mins

kordianbruck:
e voilá: http://json.tum.sexy/

PR #86, opened 2017-04-17
Allow for non TUM user use

mathiasquintero:
Sorry. I forgot about our branch-naming guidelines ;)
I'll remember next time

This solves:
https://github.com/TCA-Team/iOS/issues/64

And since I had already touched the file:
https://github.com/TCA-Team/iOS/issues/84


TUM Campus App/MoreTableViewController.swift
```diff
@@ -18,6 +25,17 @@ class MoreTableViewController: UITableViewController, ImageDownloadSubscriber, D
         }
     }
     
+    let secionsForLoggedInUsers = [0, 1]
+    let notImplemented = [
```
mammuth: Yep, that's more beautiful than my approach 👍

mathiasquintero: I just noticed I can drop the outlets for the cells ;)

PR #87, opened 2017-04-17
Add TUM Sexy

mathiasquintero:
It might not be the sexiest UI. But a feature is a feature. 

This presents all the links for TUM Sexy in a simple list.

We could do more later. But for now, it works.


TUM Campus App/TumSexyManager.swift
```diff
@@ -0,0 +1,30 @@
...
+import Sweeft
+
+class TumSexyManager: Manager {
+    
+    required init(mainManager: TumDataManager) {
+        // Nothing required
+    }
+    
+    func fetchData(_ handler: @escaping ([DataElement]) -> ()) {
+        Alamofire.request("http://json.tum.sexy").responseJSON { response in
```
mammuth: Well, it's probably a bit too strict calling this a critical/blocking issue, my friend @TCA-Bot

mathiasquintero: Well if it want's to fix it we might as well fix it

mammuth:
The icon doesn't have the correct size / padding I think.
image

Maybe it makes sense to have a static text label above the list which states what TUM.sexy is? Perhaps also with a link to the frontpage http://tum.sexy?
Otherwise the user might just not understand this view^^

TUM Campus App/TumSexyManager.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import Alamofire
+import SwiftyJSON
+import Sweeft
+
+final class TumSexyManager: Manager {
+    
+    let url: String
+    
+    init(url: String = "http://json.tum.sexy") {
```
mammuth: 🙈 😆

mathiasquintero: I don't know about you. But I count that as customizable

mammuth: It definitely is customizable

mammuth:
Icon is ok now.
If you want to add the text label, this is the phrase we're using on the website:
"No need for remembering long and ugly urls or bookmarks— Just use these redirects!"

mathiasquintero:
That was a bit too long. So I wrote "All the Redirects!". Is it ok or do you have any other preference?

mammuth:
> That was a bit too long. So I wrote "All the Redirects!". Is it ok or do you have any other preference?

Aye 👍 
Can be merged from my side.

TUM Campus App/TumSexyTableViewController.swift
```diff
@@ -0,0 +1,57 @@
...
+    
+    var entries = [SexyEntry]()
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        delegate?.dataManager().getSexyEntries(self)
+    }
+    
+    @IBAction func visit(_ sender: Any) {
+        guard let url = URL(string: "http://tum.sexy") else {
```
mathiasquintero: I know. Bot. And I don't care

mathiasquintero: Will be fixed in: https://github.com/TCA-Team/iOS/issues/90

kordianbruck:
Ok my 2 cents on this:
- I kinda think the dress is inappropriate ;)
- We should streamline features between Android & iOS otherwise I will have a bad time with support. So it would be great if you can post 2-3 Screenshots in the Android Repo if you are adding features. Just as a reminder, that this feature is not added there yet.

PR #97, opened 2017-04-19
Tim/82 card type label

TG908:
I got the Card type labels ready. If we still need this before maybe changing the card design...


mathiasquintero:
The padding at the top was really bugging me. So I fixed it. I hope you don't mind

Issue #76, opened 2017-04-11
Card showing rented books and their expiration dates


TG908:
TUM UB OPAC seems to have an api.
There is an opensource app for accessing [opac:](https://github.com/opacapp/opacclient)
Their api uses this: `"_id":"Muenchen_TU","city":"M\u00fcnchen","api":"sisis","title":"Unibibliothek TUM","account_supported":true,"_active":true,"state":"Bayern","_version_required":0,"information":"http://www.ub.tum.de/so-erreichen-sie-uns/oeffnungszeiten","data":{"baseurl":"https://opac.ub.tum.de/InfoGuideClient.tumsis","startparams":"Login=wotum"}},{"_plus_required":false,"country":"Deutschland","library_id":5606,"geo":[48.1351253,11.5819806],"`

TG908:
The TUM OPAC still uses TLS 1.0 but App Transport Security requires TLS 1.2. We could change NSExceptionMinimumTLSVersion but Apple might reject our app for that and we have to justify the use of TLS. 1.0 (we might be able to justify it becaue we have no control over the opac server)

mathiasquintero:
I doubt they will reject it for that. Apple tried to enforce the requirement for HTTPS for iOS 10.0 and waaaay too many Developers complained that they didn't have control over the APIs.

TG908:
Well then id suggest using [Kanna](https://github.com/tid-kijyun/Kanna) for scraping the html

TG908:
Got a proof of concept working (performing a login 👍)

TG908:
I am now able to query the rented books and their due date. Integration into the campus app will begin soon.

kordianbruck:
Sounds great! Eager to see how this works :dancer:

TG908:
@mathiasquintero It'll be nice if you could have a look at my manager class and my card. I might have made a mistake somewhere...

**Todo**
- [ Get Card working]
- [ Get manager working]
- [ Error handling and type safety]
- [ handle session timeouts]
- [ Prompt for password and safe password somewhere] 
- [ Improve card (add table to display all the books rented)]

Issue #43, opened 2016-12-16
Refactor use of TUMCabe news API?


mammuth:
The news API (https://tumcabe.in.tum.de/Api/news/) currently contains 3 sources:
- TUM news
- TU movie
- ~~MVG Störungsticker~~

In Android the approach looks like this as far as I understand:
- Fetch items from the news/ endpoint once
- Get the closest TUM news and display it as card
- Get closest movie news and display as card
- Get closest Störungsticker news and display as card

In iOS we currently just display the most recent news in the news card and the most recent TU movie in the movie card. However the movie is managed by the MovieManager (and it made it's own request to the kino/ API endpoint to build that card).

Maybe we should adopt the design of the Android app and build 3 cards (movie, news and MVG Störung) via one single news/ call?
The kino/ endpoint is then used in the detail view of a movie.

Related "MVV malfunction card" #25

mathiasquintero:
To make the design of the managers as easy as possible, perhaps we should introduce an enum to the news for this. Or does the model of the news vary?

mammuth:
> Or does the model of the news vary?
I guess we can keep one generic news model for all three types of news - at least that's what the API seems to be doing.

kordianbruck:
![screenshot from 2016-12-16 23-17-33](https://cloud.githubusercontent.com/assets/298860/21280376/e6b420da-c3e5-11e6-964d-a68f9c201d0f.png)

Please make this dynamic. We have more than three news sources and there will be more. The Fachschaft MPI and Maschinenbau will be added in the coming weeks.

mathiasquintero:
@kordianbruck what exactly do you mean by dynamic? My question is how does the model look like? Will it change when it's a MPI news item? (are there extra properties for those?)

kordianbruck:
No, there will not be extra fields but more sources. For each source the latest card should be displayed. Except in the tufilm, it should be the next card.

kordianbruck:
@mammuth Störungsticker proved to not be useful and was removed two months ago.

PR #98, opened 2017-04-19
Tim/76 book rental card

TG908:
#76


mathiasquintero:
This is too stateful for my taste. I'll take a look at it. How can I test it? Username and password are not set

TG908:
@mathiasquintero what do you mean with stateful? It has to be session oriented because you are basically clicking through the website in code. But it works! Please user your TUM OPAC username and password

mathiasquintero:
@TG908 I cleaned it up a bit. Took out everything that was being stored and not used again. Since we're doing a lot of similar API calls I used Sweeft instead of Alamofire for the networking and went with a promise oriented solution. I want to refactor our managers to act like this as part of https://github.com/TCA-Team/iOS/issues/43 anyway

TG908:
@mathiasquintero Thanks. Were you able to retrieve some opac data?

mathiasquintero:
@TG908 What do you mean with retrieve. I mean the requests were working for me. I don't actually have any books so it's empty for me

TG908:
@mathiasquintero that was what i meant.

mathiasquintero:
Ok. Now that's a problem. Has anyone borrowed a book recently? 
We also have to talk about how we're getting the username and password. Looks like we'll have to ask the user and persist it just like the LRZ token

TG908:
I got 2 books for testing

TUM Campus App/BookRental.swift
```diff
@@ -0,0 +1,82 @@
...
+    var cardKey: CardKey {
+        return .bookRental
+    }
+}
+
+extension BookRental {
+    
+    convenience init?(td1: XMLElement, td2: XMLElement) {
+        
+        guard let title = td1.at_css("strong")?.content,
```
mathiasquintero: @TG908 this is the the part that I think might be prone to errors. I didn't change your logic since I'm not an expert on the inner html structure

TG908: I haven't been working on code safety yet. I just wanted to create a branch with a working proof of concept as fast as possible to share the workload.

TG908: So you are processing the html in the model and not in the manager....

TG908: If you want ti see the inner html structure just `print(td1.innerHTML)`

mathiasquintero: I know it's not really MVC. But neither is iOS really. I was thinking of moving this to another file but in general, failable initializers make our work much easier and cleaner

TG908: ` print(title!)
                            print(author!)
                            print(id!)
                            print(deadline!)
                            print(bib!)
                            print(prolong!)
                            print("-----------------")`

TG908: Force unwrapping the title with 2 books on my account didnt crash the app. but I'd rather go with default values if something goes wrong.  ??

TG908: If the opac html changes all kind of things can happen

mathiasquintero: I would actually argue that we don't want to show a Book Rental if the book has no title right? That looks like there's a bug in the system. So we might as well drop the result

TG908: sure

TG908:
Where should we put the login prompt for the username and password and how should we store the password.

mathiasquintero:
We can store the password with the rest of the user details. Just have them be optional of course. As for prompting I'd suggest we ask for the login info in the detail screen for the book rentals

kordianbruck:
Is there some guideline on how to do that? I rather not have real credentials saved anywhere in the app. Some sort of encryption or something?

TG908:
@kordianbruck We can use the Keychain to secure our passwords.

kordianbruck:
> @kordianbruck We can use the Keychain to secure our passwords.

Excellent! How much more time do you guys need? I'd like to deploy today!

TG908:
Im working on the Keychain thing today. I guess ill be ready this evening.

TG908:
Keychain implementation is done. But i haven't tested it yet and it needs some more fine tuning. The login UI also has to be improved. Will do that when I come back home this evening

TG908:
I used a KeychainWrapper form some apple sample code. I hope this is not a problem.

TG908:
Also Xcode screwed up and didn't commit the wrapper

TG908:
For every book i rented, a new card appears is that right?

kordianbruck:
> For every book i rented, a new card appears is that right?
sure, why not.

mathiasquintero:
I would actually recommend only showing the most critical one. The one you have to return the soonest. Or a card listing all of them

mammuth:
A card for every book will be too much for some users. We should only use one card. 

Or a summary like 
"4 books rented
1 book to return until thursday "

This approach obviously only makes more sense than just listing the books (and maybe their date of return), if the user has a lot of rented books.

kordianbruck:
Sorry, probably makes more sense. I like never rent any books, so I just wouldn't know about those things.

TUM Campus App/TumOPACApi.swift
```diff
@@ -0,0 +1,13 @@
...
+//  TUM Campus App
+//
+//  Created by Tim Gymnich on 24.04.17.
+//  Copyright © 2017 LS1 TUM. All rights reserved.
+//
+
+import Foundation
+
+enum TumOPACApi: String {
+    case OpacURL = "https://opac.ub.tum.de/InfoGuideClient.tumsis"
```
TG908: Nope

mathiasquintero: I will clean that up later. We will move everything to a json file

TG908:
The Login screen is pretty much done still a bit crude but functional. I will be working on the Book Rental Card, when I redo all the cards.

TG908:
We still need to figure out the ATS

TG908:
If you can trust nscurl there is no way to avoid NSAllowsArbitraryLoads.

` nscurl https://opac.ub.tum.de --ats-diagnostics --verbose`

TG908:
Can we merge this soon???

PR #101, opened 2017-04-24
Tim/82 new design

### LINKED ISSUE #82: Card Design Update
What do you guys think of adding labels to the cards with the card type: e.g. Mensa, News, TU Film ...

simulator screen shot 16 04 2017 23 31 59

simulator screen

TG908:
The new Folding Cell Design
Fixes #82


TUM Campus App/CardTableViewCell.swift
```diff
@@ -7,12 +7,41 @@
...
 
 import UIKit
+import FoldingCell
 import MCSwipeTableViewCell
 
-class CardTableViewCell: MCSwipeTableViewCell {
+class CardTableViewCell: FoldingCell {
+    
     
     func setElement(_ element: DataElement) {
```
TG908: Can we replace MCSwipeTableViewCell with UITableViewCell?

mathiasquintero: MCSwipeTableViewCell is there for the swipe gestures. 
Depends on whether we don't want those anymore.

You decide. 'Cause I'm a terrible designer

TG908: It's also used for table views that don't contain any cards.

TG908: Also I can no longer use MCSwipeTableViewCell for the cards because they already inherit from FoldingCell

TG908:
Why is the emptyCard not in the cards array? This makes it hard to delete cards and requires extra effort to determine their heights.

mathiasquintero:
There's a difficult Merge Conflict. I can't fix it from here. 
Two files in the xcodeproj have the same ID. 

I suggest:
1. Remove the GradesCardCell file from the project (don't delete it). 
2. Merge Develop into your branch. 
3. Add the GradesCardCell file again

TG908:
@mathiasquintero I still need to wait for the new model. The app will crash if you try to sort the cards because the grades card and books card appear multiple times

TG908:
outdated

PR #106, opened 2017-04-28
Updating Sweeft

mathiasquintero:
### Issue
I plan on refactoring our networking layer to better suit our needs. For that, I did some changes to Sweeft that I wanna work with.

My plan:
- Refactor Networking to remove as much copy and pasted code as possible.
- Remove all the hardcoded urls and move them to a json file we can change more easily
- Add Persisting Caching Layer

Updated to the latest in Sweeft


mathiasquintero:
Oops I tinkered with the pods checksum a bit. Didn't think it would trip on the pods version. I will fix it

mammuth:
However, the Travis build seems to cannot resolve Sweeft 0.8.0 (yet)

mathiasquintero:
Yet!

mathiasquintero:
@mammuth What do you think of my idea for refactoring so far.

So something like the CalendarManager would look like this:

```
final class CalendarManager: CachedManager, SingleItemManager {
    
    typealias DataType = CalendarRow
    
    var config: Config
    
    var cache = [CalendarRow]()
    var isLoaded = false
    
    var requiresLogin: Bool {
        return false
    }
    
    init(config: Config) {
        self.config = config
    }
    
    func toSingle(from items: [CalendarRow]) -> DataElement? {
        let new = items |> { $0.dtstart! > .now } // TODO:
        return new.first
    }
    
    func perform() -> Response<[CalendarRow]> {
        return config.tumOnline.doXMLObjectsRequest(to: .calendar, at: "events", "event")
    }
    
}
```

TG908:
@mathiasquintero  Looks very clean. Where would you implement the caching logic?

mammuth:
Nice!

mathiasquintero:
The logic would be part of the CachedManager protocol. 

I already started on a branch if you wanna check it. I'm not too sure how I will serialize and deserialize the data. Intuitively I should store it in the format it came in. But since we have 3 formats: JSON, XML and HTML, it won't be so pretty ;)

TG908:
@mathiasquintero maybe nskeyedarchiver??

mathiasquintero:
@TG908 That's just an extra step. I don't want to have to specify the encoding and decoding again. If I simply store the response as is and do some clever extensions depending on the init of the type maybe we can just feed it the same response again

kordianbruck:
So this is also ready then? :dancer:

PR #114, opened 2017-05-02
Tim/108-emailFeedback

TG908:
Added subject: [iOS] to support mail

### DIFF
--- a/TUM Campus App/MoreTableViewController.swift
+++ b/TUM Campus App/MoreTableViewController.swift
@@ -8,8 +8,9 @@
 
 import UIKit
 import Sweeft
+import MessageUI
 
-class MoreTableViewController: UITableViewController, ImageDownloadSubscriber, DetailViewDelegate {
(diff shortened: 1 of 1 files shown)


kordianbruck:
Can you also add the app version and ios version to the body please?

http://www.rapidtables.com/web/html/mailto.htm

TG908:
How to you want it exactly: So far ive got
Subject: [iOS: 10.3.1, App Version 1.1]

mammuth:
Nah, don't pack everything into the subject. You should be able to provide a text for the mail body as well 😉 Just keep [iOS] in the subject and add the versions to the mail body. 

It should be as easy as something like
_mailto:tca-support.os.in@tum.de?subject=[iOS]&body=iOS%20version:%20\(systemVersion),%20App%20Version:%20\(appVersion)_. 
At least on the web it is as easy 😛 

You probably need to urlencode it to remove whitespaces etc.

TG908:
when i use mailto with a longer string the mail app will not open at all. Can but the version in the body though

mathiasquintero:
@TG908 Check out `MailComposeController`

TG908:
I am already using MFMailComposeViewController

PR #120, opened 2017-05-05
Tim/table view fixes

### LINKED ISSUE #70: Problems / Bugs from Userfeedback
- [x] Unter more ganz oben wo anfangs der Name stand, steht jetzt nichts mehr #118
- [x] Klicke ich auf das Defaultbild im Kreis stürzt die App ab #118
- [ ] Noten, Tuition und Vorlesungen werden nicht mehr angezeigt (bei der ersten Benutzung der App wurden diese noch angezeigt)
- [x] Service und Default Campus lassen sich nicht öffnen #80

### LINKED ISSUE #118: More View crashes/doesnt display name

```
var user: User? {
        didSet {
            updateView()
        }
    }
```

Related to #70 

TG908:
### Fixes

- About Cell is getting deselected after selection
- navigationItem title for grades
- Fixed crash when viewing PersonalDetail on bad internet connection
- User with bad internet connection is treated as not logged in

**For this to work properly the model needs to notify the vc when user.data was set!**

fixes #70, fixes #118

### DIFF
--- a/TUM Campus App/GradesTableViewController.swift
+++ b/TUM Campus App/GradesTableViewController.swift
@@ -32,6 +32,7 @@ extension GradesTableViewController {
         delegate?.dataManager().getGrades(self)
         tableView.estimatedRowHeight = tableView.rowHeight
         tableView.rowHeight = UITableViewAutomaticDimension
+        navigationItem.title = "Grades"
     }
     
 }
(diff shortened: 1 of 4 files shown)


TG908:
implemented in #192 and outdated in parts

Issue #124, opened 2017-05-10
Picker views don't hide automatically


mammuth:
The picker views don't hide automatically when they are open and the user navigates back. This occurs at least with the following views (add more when you find/remember them):
- Cafeteria Chooser
- Study Room Buildings
- Room Finder 
- ...

Related support tickets (tell them via mail when this is fixed):
- http://tca.exgen.io/scp/tickets.php?id=209

TG908:
Also the arrow doesn't update when you dismiss the picker by tapping anywhere

PR #126, opened 2017-05-10
Dismiss picker on view disappearing

mammuth:
#124

### DIFF
--- a/TUM Campus App/CafeteriaViewController.swift
+++ b/TUM Campus App/CafeteriaViewController.swift
@@ -91,12 +91,15 @@ extension CafeteriaViewController {
         pickerView.show()
         barItem?.action = #selector(CafeteriaViewController.hideCafeterias(_:))
         barItem?.image = UIImage(named: "collapse")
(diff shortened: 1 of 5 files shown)


mammuth:
When using this approach, the target-view isn't positioned properly anymore. Any guess why @mathiasquintero ?

Is there maybe a simpler solution than adding this to every VC?

image

TG908:
I fixed it for you hope that is ok. I no longer allow the user to tap the back button when the slider is visible.

TG908:
Damn it fixed hitting the back button but sliding on the edge of the view to go back still causes the glitch...

mathiasquintero:
try disabling user interaction in the navigation controller

mammuth:
Yep, `isUserInteractionEnabled` works 👍

mammuth:
Argh, but the way I did it, of course also blocks clicking the back button or picker button...

mammuth:
Now it should behave as intended I think 😊

mammuth:
The picker is pretty duplicated in 5 VC's, maybe we get this in and make an issue for refactoring it sometime?

Issue #103, opened 2017-04-26
Calendar Details (exact time and room)


kordianbruck:
"Im Calendar wäre es gut, wenn man die genauen Zeiten sieht, sobald man auf eine Veranstaltung klickt. Oder dann vielleicht zu den Veranstaltungsinformationen weitergeleitet wird. "

TG908:
and the room

mathiasquintero:
Then we could listen for URL calls and intercept when someone opens a link and show it in the app

mammuth:
Folgende User per Mail informieren:
- https://tca.exgen.io/scp/tickets.php?id=211
- https://tca.exgen.io/scp/tickets.php?id=195
- https://tca.exgen.io/scp/tickets.php?id=206

mathiasquintero:
It would be great if we could display the room finder. However I'm having trouble finding the roomId from the calendar events location

kordianbruck:
@mathiasquintero you will have to search for the room name and let the user chose. Thats what we do in Android. There is unfortunately no exact matching.

PR #129, opened 2017-05-11
Tim/103 calendar location

TG908:
### Issue

Added location to nextLecture card and calendar view

### This fixes the following issue(s):
#103


TG908:
Since reworking the model seems to take a bit longer... I'd suggest releasing version 1.1.5 this week with the location, the sliderView fix and the email subject.

Issue #139, opened 2017-08-12
Login flow needs improvement


lukaskollmer:
**Problem:**

Selecting "I don't have an LRZ-ID" in the login screen will load the "Main" storyboard and present the `CampusTabBarController` on top of the login view controller.
This works fine the first time, but due to how the login screen is presented (reassigning `UIApplication.shared.keyWindow?.rootViewController` rather than simply calling `UIViewController.present(_:animated:completion:)`), a new instance of `CampusTabBarController` is created every time you skip login.
This means that - if you open and dismiss the login screen a couple of times - the app's view hierarchy gets unnecessarily convoluted, which in turn consumes a lot of memory:

![](https://files.lukaskollmer.me/embed/tum-campus-app-bad-view-hierarchy.png?v=2)

**Solution:**

The obvious fix is to
1) Re-implement the login flow and present the `LoginViewController` by calling `present(_:animated:completion:)` on the `CampusTabBarController`
2) Simply call `dismiss(animated:completion:)` on the `LoginViewController` when the skip button is pressed
3) Use singletons (loading all view controllers from their storyboards every time you're about to present them is an incredibly inefficient approach)

TG908:
You‘re right would you mind putting this into a PR soon?

lukaskollmer:
I'll see what I can do but I'm new to the codebase so this might take a while

mobi-robert:
Hi, I'd like to implement this, if that's alright.

TG908:
Sure go ahead.

mammuth:
Closed via #143

PR #143, opened 2017-10-06
139 - Fix issue where the login view controller was presented inefficiently

rodalfus:
### Issue
- This pull request fixes an issue in which the login view controller would be loaded from the storyboard every time it was opened
- This pull request fixes an issue in which the login view controller would be set as `window.rootViewController` and therefore many copies of it would be kept in memory if it was opened and closed often.
- This pull request fixes a spelling issue with the title of the view controller that is presented when awaiting token access.

#139 

### Screenshots
screen shot 2017-10-06 at 17 45 42
This is after opening and closing the login view controller about five times.


rodalfus:
@TG908 Is there any way you could restart the build? The logs suggest this might be a one-time failure and I definitely did not change anything regarding the CI :) 

Relevant logs:
```
$ pod repo update > /dev/null
  error: RPC failed; curl 56 SSLRead() return error -36
  fatal: The remote end hung up unexpectedly
  fatal: early EOF
  fatal: index-pack failed
The command "pod repo update > /dev/null" failed and exited with 1 during .
```

mammuth:
@rodalfus I restarted the CI job.

rodalfus:
Huh. That is weird. Have you ever encountered this before @mammuth?

mammuth:
Strange. We need to investigate on this.

mathiasquintero:
I think we can remove the update all together for now. 
This is there because I sometimes develop functions I need here on other projects and add them as deps.
I'm sure all the pod caches have the needed versions.

TUM Campus App/ViewControllerProvider.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+//  Created by Robert on 06.10.17.
+//  Copyright © 2017 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+
+class ViewControllerProvider {
+    static private(set) var loginNavigationViewController: UINavigationController? = {
+        let storyboard = UIStoryboard(name: "Setup", bundle: nil)
```
mathiasquintero: I'm still not a big fan of adding more Singletons. But what the hell, who cares?!?

Podfile.lock
```diff
@@ -59,4 +59,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 6ee09db13e81dfe3abdf89ad811b4de542da1a35
 
-COCOAPODS: 1.2.1
+COCOAPODS: 1.3.1
```
mathiasquintero: Are we all on this version?

TUM Campus App/PersonDetailTableViewController.swift
```diff
@@ -93,13 +93,20 @@ extension PersonDetailTableViewController {
             }
         }
         let cell = tableView.dequeueReusableCell(withIdentifier: "contact") ?? UITableViewCell()
-        cell.textLabel?.text = contactInfo[indexPath.row].0.rawValue
-        cell.detailTextLabel?.text = contactInfo[indexPath.row].1
+        if indexPath.row < contactInfo.count{
```
mathiasquintero: Please adhere to the rough coding guidelines we have. Space before the `{` and `else` in the same line `}`. For the sake of consitency

TUM Campus App/AppDelegate.swift
```diff
@@ -15,42 +15,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
-        //UIApplication.sharedApplication().setStatusBarStyle(UIStatusBarStyle.LightContent, animated: true)
+        setupAppearance()
+        conditionallyShowLoginViewController()
+
+        return true
+    }
+
+    private func setupAppearance(){
```
mathiasquintero: Please fix the spacing here as well. I know it's tedious, but my OCD and my internal linter are going crazy when I see this

TUM Campus App.xcodeproj/project.pbxproj
```diff
@@ -828,6 +831,14 @@
 			path = AutomatedScreenshots;
 			sourceTree = "<group>";
 		};
+		E1369C9B1F87D22200D6766E /* UI Services */ = {
```
mathiasquintero: New Group? Nice!

TUM Campus App/ViewControllerProvider.swift
rodalfus: To be honest, me neither. However, I guess in this case it is fine, since we just kind of "bundle" the work that other parts of the code might have done in one place that is easy to get to. I guess in this case there is no really strong argument for or against either way.

Podfile.lock
rodalfus: Huh. I did not notice that being updated. Can revert, if it's supposed to be `1.2.1`.

TUM Campus App/ViewControllerProvider.swift
```diff
@@ -0,0 +1,18 @@
...
+//  TUM Campus App
+//
+//  Created by Robert on 06.10.17.
+//  Copyright © 2017 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+
+class ViewControllerProvider {
+    static private(set) var loginNavigationViewController: UINavigationController? = {
```
mathiasquintero: Do you absolutely need it to be of type `UINavigationController`? Otherwise, change it to `UIViewController` so that it isn't optional and then you can get rid of the guard with break inside a switch

TUM Campus App/ViewControllerProvider.swift
```diff
@@ -0,0 +1,18 @@
...
+//  ViewControllerProvider.swift
+//  TUM Campus App
+//
+//  Created by Robert on 06.10.17.
+//  Copyright © 2017 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+
+class ViewControllerProvider {
```
mathiasquintero: I don't know about this. Perhaps we should have a protocol for this and call it always directly on the class.
e.g.: `protocol InstantiableByStoryboard { static var shared: Self { get } }`

Don't worry too much about this. It's just a thought for later ;)

Podfile.lock
mathiasquintero: Not a problem. I just hate it when every PR changes the pods version ;)

rodalfus:
Perhaps we need to add a step to update fastlane during the build on Travis. Not sure why the —force flag is not available. Will look into this tomorrow

.travis.yml
```diff
@@ -16,7 +16,8 @@ before_install:
   - ./fastlane/before_install.sh
 
 script:
-   - fastlane snapshot reset_simulators --force  # Workaround for Travis bug https://github.com/travis-ci/travis-ci/issues/7031
+   - export SNAPSHOT_FORCE_DELETE=true
+   - fastlane snapshot reset_simulators  # Workaround for Travis bug https://github.com/travis-ci/travis-ci/issues/7031
```
mathiasquintero: @rodalfus I'm pretty sure the problem with the build is this line: https://github.com/TCA-Team/iOS/blob/master/.travis.yml#L15

mathiasquintero: Oops nevermind. Looked at the latest log. That wasn't the problem. Weird...

rodalfus: @mathiasquintero Actually, you might be right. Looking at the running build, this error message is created:
```
Fetching: public_suffix-3.0.0.gem (100%)
ERROR:  Error installing fastlane:
	public_suffix requires Ruby version >= 2.1.
```
Perhaps that is the reason that a version of fastlane is used that does not support the `--force` option on `reset_simulators` since it's too old.

rodalfus: Yup, that seems to be it.
```
[!] The Fastfile requires a fastlane version of >= 2.19.1. You are on 1.111.0.
```

mathiasquintero: Perhaps the node using an outdated image. Maybe try using Xcode8.3

mathiasquintero: @mammuth any idea why the tests might be failing. They are complaining about us using a deprecated API but I'm not sure that's it

rodalfus:
I'll fix this warning and then I think we're good to go
```
WARNING: output_types and output_files do not have the same number of items. Default values will be substituted as needed.
```

mammuth:
The reason why we reset the simulators was [this bug](https://github.com/travis-ci/travis-ci/issues/7031) with Travis' Xcode8.2 image.
Apparently in the 8.3 image this is no problem anymore (but they now have duplicate watchOS simulators... 🤦‍♂️ ), so we might test to just remove the `reset_simulators` to get faster builds 🚀 

> Perhaps we need to add a step to update fastlane during the build on Travis.

Well, we install fastlane [ourselves](https://github.com/TCA-Team/iOS/blob/master/fastlane/before_install.sh#L3) and pinned it to that specific version which definitely worked in the past.

Btw, so if we update fastlane, we need to do it in the CI install script AND change the version pinning/check in the Fastfile.

rodalfus:
> so we might test to just remove the reset_simulators to get faster builds
@mammuth This seems a little out of scope for this pull request. We could (I'll gladly do that) try that out in another one though (for the sake of keeping everything organised and _on topic_).

mammuth:
> @mammuth This seems a little out of scope for this pull request. We could (I'll gladly do that) try that out in another one though (for the sake of keeping everything organised and on topic).

Sure, go for it.  I just wanted to drop the info here since it was related 😊

Issue #145, opened 2017-10-07
Check if resetting the simulators is still needed for the builds to function correctly on Travis


rodalfus:
The reason why we reset the simulators was this bug with Travis' Xcode8.2 image.
Apparently in the 8.3 image this is not a problem anymore, so we it should be tested if just removing the reset_simulators results in faster builds.

rodalfus:
I'll start work on this.

PR #146, opened 2017-10-07
#145 Remove `reset_simulators` build step

rodalfus:
### Issue
In Xcode-8.2 it was necessary to remove the simulators before running the test suite because of some simulators being duplicated. 
In Xcode-8.3 this bug seems to be fixed. Therefore our build times should decrease

#145

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ osx_image: xcode8.3
 
 branches:
   # Only run push builds for the master branch. PR builds are run nevertheless
-  only: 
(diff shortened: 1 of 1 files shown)


rodalfus:
Please do **not** merge this yet. We should examine the Travis results as well as build times first.

rodalfus:
@mammuth Ok, so it does not seem to make a _huge_ difference from a build time point of view, but I still believe that the removed complexity makes it worth it.

Issue #151, opened 2017-10-15
Pasting into the login ID textfields is wonky


rodalfus:
**Steps to reproduce**:
(1):
- Open login screen
- copy your ID (e.g. ga12ab)
- paste the ID in the first textfield

You will see that the whole ID is pasted into one textfield.

(2):
- Open login screen
- Input the first two characters of your ID and then select the first textfield again
- Input another character

You will see that it is possible to input more than the expected two characters into the textfield.

**Definition of Done**:
- Every textfield has atmost the specified number of characters in its text
- Pasting into one textfield can "overflow" text to the next ones

rodalfus:
@mathiasquintero I believe that a solution that does not require third-party libs and works as expected (as I believe the one I implemented does) is preferable to using a third-party (especially) UI component, which we'd have to make sure is up-to-date and might just be abandoned in a year or so.

mathiasquintero:
Actually I was thinking that maybe at this point UIKit may have a solution

rodalfus:
I'll have a look, but honestly I doubt that there is a native UIKit solution for this. Have not come across one until now at least ;)

PR #152, opened 2017-10-15
151 - Fix issues with wonky ID input

rodalfus:
### Issue

**Steps to reproduce**:
(1):

Open login screen
copy your ID (e.g. ga12ab)
paste the ID in the first textfield
You will see that the whole ID is pasted into one textfield.

(2):

Open login screen
Input the first two characters of your ID and then select the first textfield again
Input another character
You will see that it is possible to input more than the expected two characters into the textfield.

**How this fixes the issue**:

- Specify number of characters fitting into each textfield explicitly
- Use the replacement string's indexes to find out where to put a given character

#151

### DIFF
--- a/TUM Campus App/LoginViewController.swift
+++ b/TUM Campus App/LoginViewController.swift
@@ -76,8 +76,12 @@ extension LoginViewController: UITextFieldDelegate, TokenFetcherControllerDelega
         
         for i in textFieldIndex..<textFields.endIndex {
             let startIndex = mutableReplaced.startIndex
-            let endIndex = replaced.index(startIndex, offsetBy: min(mutableReplaced.characters.count, numberOfCharacterPerTextField[i]) - 1)
+            let endIndex = replaced.index(startIndex,
+                                          offsetBy: min(mutableReplaced.characters.count,
+                                                        numberOfCharacterPerTextField[i]) - 1)
+            
             let range = ClosedRange(uncheckedBounds: (lower: startIndex, upper: endIndex))
+            
             textFields[i]?.text = mutableReplaced[range]
             textFields[i]?.becomeFirstResponder()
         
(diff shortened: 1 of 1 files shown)


TUM Campus App/LoginViewController.swift
```diff
@@ -58,39 +58,53 @@ extension LoginViewController: UITextFieldDelegate, TokenFetcherControllerDelega
...
+        if textFieldContentsAreValid() {
+            self.view.endEditing(true)
+            performSegue(withIdentifier: "waitForConfirmation", sender: self)
+        }
+        
+        return false
+    }
+    
+    private func textFieldContentsAreValid() -> Bool {
+        guard let firstText = firstTextField.text, let numbers = numbersTextField.text, let secondText = secondTextField.text, let _ = Int(numbers)
```
mathiasquintero: Please have these in separate lines. Just a new line after each comma and then the else block also in a new line

TUM Campus App/LoginViewController.swift
```diff
@@ -58,39 +58,53 @@ extension LoginViewController: UITextFieldDelegate, TokenFetcherControllerDelega
...
+            return false
+        } else if string.characters.count < range.length {
+            return true
+        }
+    
+        var mutableReplaced = replaced
+        
+        for i in textFieldIndex..<textFields.endIndex {
+            let startIndex = mutableReplaced.startIndex
+            let endIndex = replaced.index(startIndex, offsetBy: min(mutableReplaced.characters.count, numberOfCharacterPerTextField[i]) - 1)
```
mathiasquintero: put each argument in a new line. As a rule of thumb. If I can't read it without scrolling maybe a new line would help

TUM Campus App/LoginViewController.swift
```diff
@@ -58,39 +58,53 @@ extension LoginViewController: UITextFieldDelegate, TokenFetcherControllerDelega
...
+            let range = ClosedRange(uncheckedBounds: (lower: startIndex, upper: endIndex))
```
mathiasquintero: Uhhh! Unchecked Bounds, scary stuff... ;) Are you sure about this. I understand that you don't want the range to throw a fatalError if endIndex < startIndex.

Care to elaborate on why that happens?

rodalfus: Hmm, what would be the alternative to using `uncheckedBounds`? The `endIndex < startIndex` would only ever be true if one of the `numberOfCharactersPerTextField` were to be `0` or less. Perhaps I should be explicit about that being an array of unsigned integers.

mathiasquintero: `(startIndex..<endIndex)` should also work, but this is fine

PR #153, opened 2017-10-16
updated pods

TG908:
### Issue


mammuth:
Introduces a couple of errors, let's track that together with the Swift 4 transition

Issue #158, opened 2017-10-18
Calendar does not show any events


mammuth:
I think we don't have an issue for that bug yet. There are many reports about this in the ticket system. They state they are able to see their lectures and stuff like that, but their calendar is empty.

mathiasquintero:
I wouldn't necessarily call this a bug.
We need to add a button to force refresh the calendar, instead of the user having to wait a week for it to refresh

mammuth:
So it's due to the caching in the app? Then I'd still call it a bug if the cache period is a week, since that's no proper cache rate in my opinion.

There are also un-periodic events the user is not aware of, which are placed spontaneously into the calendar by TUM Online. Receiving this event one week late might be bad as well...

kordianbruck:
@mammuth we have to respect TUMonline here. Fetching the calender is one of the most impacting operations and requires a lot of computation time on their end. One week is resonable as mostly the lectures change once at the beginning and then some get canceled. But almost never are there any lectures added short hand.

mammuth:
Wut, TUM Online wants us to have a calendar that's in the worst case one week out of date 😮 

But okay, refresh button it is...

mammuth:
Resolved by #162

PR #162, opened 2017-10-19
Add refresh capabilities to calendar

mammuth:
@mathiasquintero could you please review this first draft? 😊 

I used a custom button because I thought this way it'll be easier to animate it. I'm not sure whether it's also easy doing it with a BarButton so we could use UIBarButtonSystemItem.refresh as an image.

**ToDo**

- [x] Use proper Icon
- [x] Signal update is in process, eg. by rotating the icon until the handler is called

#158

### DIFF
--- a/TUM Campus App/CalendarManager.swift
+++ b/TUM Campus App/CalendarManager.swift
@@ -50,6 +50,11 @@ class CalendarManager: Manager {
         }
     }
     
+    func updateData(_ handler: @escaping ([DataElement]) -> ()) {
+        print("Updating calendar data...")
+        fetchFromTUMOnline(handler)
+    }
+    
     func getDocumentDirectory() -> String {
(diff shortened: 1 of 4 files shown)


mathiasquintero:
I like the idea. Perhaps we should include that in my current refactoring. Since I want every module expect for search to be cached. Perhaps we just need a `byPassCache` option flag any managers fetch operation.

mammuth:
You mean a bypass parameter in the fetchData method? Would be an optoin aswell 👍 
An extra updateData like I did may be more flexible.

mammuth:
Now I'm replacing the refresh system icon with the system stop icon to indicate a refresh is in process. 

This PR is ready from my side now!
image

mammuth:
I'd say let's get this in so we have it for 1.2 and change it afterwards based on @mathiasquintero refactorings in #167.

Wainting for review @mathiasquintero, @TG908 and others 😊

Issue #150, opened 2017-10-09
Token not correctly propagated to managers after login


rodalfus:
Steps to reproduce:
- Log out of application if logged in
- Log in to application
- Go to "More" tab
- Check if your name is displayed (it won't be)
- Force quit and restart application
- Check if name is displayed in "More" tab (it will be)

This issues seems to been introduced with #143 (sorry!). Furthermore, the logs suggest that the token used by the application to get the user is incorrect until forcing a restart. Perhaps the reason for this can be found in one of the `viewDidLoad`s or `init`s of a view controller that was previously reinitialised (just trying to reason about this - might be something totally different).

This issue should be fixed before a new version is uploaded to the app store.

rodalfus:
Ok, so I think we can fix this in the following way: add a reference to the `TUMDataManager` used in the `MoreTableViewController` to the `WaitForTokenViewController`. Update the `dataManager`'s user once we get a new one in the `MoreTableViewController`. Also, request the user's user data there. Add a callback to `TUMDataManager.getUserData` and update the `MoreTableViewController`'s views when that is called. 

Am I missing anything here? Does anyone have a better idea?

rodalfus:
@mathiasquintero

mathiasquintero:
Add a delegate to the MoreTableViewController and then you'll have a reference to DataManager through the delegate. I don't know about updating the data everytime, though. Maybe only if it's nil

mammuth:
Resolved by #170

PR #170, opened 2017-10-20
150 - Fix issue with user info not being updated correctly

rodalfus:
### Issue
When logging in, the user's name and data would not be updated correctly. The reason for this was that the token used to update the user's name was not forwarded to the `TUMDataManager`. This pull request fixes that in the following way: set the `TUMDataManager`'s user to be the shared (updated) user, add a fetch data method with a callback to `TUMDataManager`, thereby allowing the `MoreTableViewController` to update its data once the user is logged in.

#150 

If there is no significant difference between the user object in `User.shared` and the user object in `TUMDataManager`, we could combine the two (with the added benefit, that the card view controller would work fine out of the box as well).

### DIFF
--- a/TUM Campus App/CampusTabBarController.swift
+++ b/TUM Campus App/CampusTabBarController.swift
@@ -17,7 +17,7 @@ class CampusTabBarController: UITabBarController {
         UITabBar.appearance().tintColor = Constants.tumBlue
         UITabBar.appearance().backgroundColor = UIColor.white
         UITabBar.appearance().barTintColor = UIColor.white
-        manager = TumDataManager(user: User.shared)
+        manager = TumDataManager()
         manager?.getUserData()
     }
 
--- a/TUM Campus App/CardViewController.swift
+++ b/TUM Campus App/CardViewController.swift
@@ -37,7 +37,7 @@ extension CardViewController: ImageDownloadSubscriber, DetailViewDelegate {
     }
     
     func dataManager() -> TumDataManager {
-        return manager ?? TumDataManager(user: nil)
+        return manager ?? TumDataManager()
     }
 }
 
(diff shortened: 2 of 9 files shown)


rodalfus:
@mammuth Is that 👍 a yes to "let `TUMDataManager.user` be `User.shared`"? In how far does this need to change with @mathiasquintero's recent changes?

mammuth:
> @mammuth Is that 👍 a yes to "let TUMDataManager.user be User.shared"?

It's just a "cool, thanks for the contribution!" 😉 

It's @mathiasquintero who is better suited for answering your second question.

mathiasquintero:
@rodalfus  Sounds good to me. One singleton less ;)

mathiasquintero:
@rodalfus can we get this one done so that it can go on the 1.2 Release

rodalfus:
@mathiasquintero Sure.

rodalfus:
@mathiasquintero Just to avoid confusion:

This is how I would go about this:

```
    var user: User? {
        return User.shared
    }
```
and remove the designated initialiser of `TUMDataManager`.

Do you have a better suggestion? Perhaps refactoring away `User.shared` is somewhat out-of-scope for a release this week (seems to be a pretty substantial change that requires extensive testing)?

rodalfus:
I manually tested the changes in the following ways:
- Log in and out "above" the More screen -> Name was updated properly
- Log in and out "above" the Cards screen, switch to More screen -> Name was updated properly (after a short delay because of the network request)
- Log in and out "above" the Cards screen, refresh cards -> user specific cards are shown
- Test Calendar, My Lectures, Tuition Fees, Study Rooms, Cafeterias, News screens (could _not_ test some of the other screens since I don't have data for those or could not install the application on my device) -> all worked as expected.

From my side, this PR is ready.

Issue #123, opened 2017-05-09
Login issues


mammuth:
There are several users who complain not being able to get to the cards view after clicking authorization after enabling the tokens.

mammuth:
Related tickets:
http://tca.exgen.io/scp/tickets.php?id=204
http://tca.exgen.io/scp/tickets.php?id=207
(Send them mails when we fixed this)

mammuth:
We should try to fix that asap since we already got a couple of more mails now that the semester starts.

@mathiasquintero If I remember correctly you told me once you found the cause for this? Can't remember what it was 🙈

mathiasquintero:
Hahaha me neither. Damn. We'll have to check it out

kordianbruck:
Wasn't this the problem that we just fixed in that pull request?  #143

mammuth:
That's why I was asking for the cause. :blush:

Wasn't there the problem that the token was requested whenever the login vc was opened or something like that? So when people activated one token in TUMOnline and came back to the app, it sometimes already had requested another one? Or is that not the current behavior anymore / got fixed?

mammuth:
@mathiasquintero could you dig into this during this week so we can release a new version at the weekend? We get a lot of support mails which are caused by our login flow.

mammuth:
Related / Similar changes: https://github.com/TCA-Team/iOS/issues/131

mammuth:
Resolved via #171

PR #171, opened 2017-10-20
Mathias/persist token awaiting

mathiasquintero:
### Issue
https://github.com/TCA-Team/iOS/issues/123

The UI is a bit funky.
Maybe @TG908 or @rodalfus can help.

Just fyi the button looks weird when we have already requested a token and awaiting confirmation

### DIFF
--- a/TUM Campus App/PersistentUserData.swift
+++ b/TUM Campus App/PersistentUserData.swift
@@ -22,11 +22,14 @@ enum PersistendUserData: Codable {
     }
     
     case no
+    case requestingToken(lrzID: String)
     case some(lrzID: String, token: String, state: State)
     
     var user: User? {
         switch self {
         case .no:
+            return nil
+        case .requestingToken:
             return nil
         case .some(let lrzID, let token, let state):
             return state == .loggedIn ? User(lrzID: lrzID, token: token) : nil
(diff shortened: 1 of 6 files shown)


rodalfus:
@mathiasquintero I guess the issue is that we started using a shared instance of the login view controller flow with #139 (underlying issue) / #143 (PR). That means that the button gets animated to be “small” / gets reset to that smaller position and seems to be kept in memory. What we could do about this is: 1) stop keeping the login view controllers in memory 2) reset the button to its initial position after the animation is done.

TUM Campus App/PersistentUserData.swift
```diff
@@ -10,20 +10,29 @@ import Sweeft
 
 enum PersistendUserData: Codable {
     
+    enum State: Int, Codable {
+        case loggedIn
+        case awaitingConfirmation
+    }
```
mammuth: 👍

TUM Campus App/WaitForTokenViewController.swift
```diff
@@ -42,30 +36,33 @@ extension WaitForTokenViewController {
...
+        } else {
+            button.startLoadingAnimation()
+            loginManager.fetch()
+        }
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        guard !PersistentUser.isLoggedIn else { return }
+        PersistentUser.reset()
```
mammuth: So when I login, get to the WaitForToken Screen and press the Home button (without authorising the token yet), the PersistentUser will get reset? Shouldn't the state still be .awaitingConfirmation ?

mammuth:
You could also think about merging master in order to get the CI up and getting feedback by our bot (if you dare showing your code to him 🙈😆)

TUM Campus App/WaitForTokenViewController.swift
mathiasquintero: Ah, I forgot. Does `viewWillDisapper` get called when pressing the home button?

mathiasquintero: I checked. I doesn't get called

mammuth: Alright! 😊

PR #178, opened 2017-10-20
New UI Design (step 4)

### LINKED ISSUE #68: Feature Proposal: New Card for Newest Grades
How about a card that shows your newest Grade? I think it would be quite handy to just take a look at your TUM app and see your newest grade when you receive the email, instead of checking out the website - Or you can see all your grades at once for the semester you're in or something.

TG908:
Goal of this step is to add a few new Cards:

- [ ] Grades
- [ ] Study Rooms
- [ ] MVV
- [ ] My Lectures

Fixes #68 ?


TG908:
Will be implemented in #209

PR #192, opened 2017-10-22
Remove TabBar

### LINKED ISSUE #118: More View crashes/doesnt display name

### LINKED ISSUE #179: Remove TabBar
As suggested by @TG908, the tabbar is kind of unneeded. 

Possible other ways to get to the more view:
- We could add the profile image to the toolbar which links to it
- some kind of swipe / dropdown similar to snapschat ( @mathiasquintero  idea)
- Your sexy idea here

### LINKED ISSUE #190: Large Titles?
What is your opinion on the large titles from iOS 11. Should we include them.

I think we should to keep our app's look as close as possible to other ios 11 apps. **With one exception: CardTableView** since we don't have a proper title for this vc (maybe someone of you guys knows a way to make the image move up like text).

TG908:
Goal of this PR is to replace the TabBar

Thing to do:

- [x] Implement new SearchController
- [x] Remove old search icon
- [x] Delete old search controller
- [x] Fix navigation issue when clicking on search result
- [x] Remove tab bar
- [x] Add Account icon
- [ ] Clean up more page
- [x] Implement large titles
- [x]  Fix large title for card view (image)

Fixes #190
Fixes #179 
Fixes #118


mathiasquintero:
@TG908 I fixed the issue with the missing Data in the screens. It was treating every user as a stranger

TUM Campus App/CardViewController.swift
```diff
@@ -72,13 +75,30 @@ extension CardViewController {
...
         tableView.separatorStyle = .none
         tableView.backgroundColor = .white
-        manager = (self.tabBarController as? CampusTabBarController)?.manager
+      
+        let storyboard = UIStoryboard(name: "CardView", bundle: nil)
+        guard let searchResultsController = storyboard.instantiateViewController(withIdentifier: "SearchResultsController") as? UIViewController else {
+            fatalError("Unable to instatiate a SearchResultsViewController from the storyboard.")
+        }
+        search = UISearchController(searchResultsController: searchResultsController)
+        search?.searchResultsUpdater = self
```
mathiasquintero: I'd suggest that the search controller should also be the updater. That way we don't have to keep a reference to it...

TG908: The search results table somehow becomes more laggy when the updater is the searchResultsController. I don't mind the reference if it gives us better performance.

TG908:
I want to remove all the options from the account (more) view that are, accessible from the cards view. Some examples are: Room Finder (is included in search bar also doesn't work anymore), Cards (doesn't reload the cards view)

mathiasquintero:
@TG908 what's the status on this?
When can we merge this?

TG908:
@mathiasquintero we are missing the accounts bar button item. We should also remove a few of the duplicate entries from the More tab

TUM Campus App/CardViewController.swift
```diff
@@ -72,21 +75,40 @@ extension CardViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         setupLogo()
+        
+        if #available(iOS 11.0, *) {
+            self.navigationController?.navigationBar.prefersLargeTitles = false
+            self.navigationController?.navigationItem.largeTitleDisplayMode = .never
```
mathiasquintero: shouldn't `largeTitleDisplayMode = .never` be enough? I think you don't have to set `prefersLargeTitles` every time

TUM Campus App/CardViewController.swift
```diff
@@ -128,3 +150,18 @@ extension CardViewController {
     
 }
 
+
+extension CardViewController: UISearchResultsUpdating {
+    
+    func updateSearchResults(for searchController: UISearchController) {
```
mathiasquintero: I think the search results controller should handle this.
And then the cards view controller is less clutered

TUM Campus App/CardViewController.swift
```diff
@@ -72,21 +75,40 @@ extension CardViewController {
...
+            fatalError("Unable to instatiate a SearchResultsViewController from the storyboard.")
+        }
+        search = UISearchController(searchResultsController: searchResultsController)
+        search?.searchResultsUpdater = self
+        search?.searchBar.placeholder = "Search"
+        search?.obscuresBackgroundDuringPresentation = true
+        search?.hidesNavigationBarDuringPresentation = true
+        if #available(iOS 11.0, *) {
+            self.navigationItem.searchController = search
+        }
```
mathiasquintero: People with iOS 10 should probably still be able to search. 
We should default to an extra search view controller here.

TUM Campus App/CardViewController.swift
TG908: As I already said, doing it in the ResultsVC makes the search laggy.

mathiasquintero: Ok I'll check it out

TG908: I think so too

mathiasquintero:
Closing since this is in #202

PR #199, opened 2017-10-26
Release/1.3

TG908:
Release 1.3 integration branch


mathiasquintero:
I still don't see why we need an extra branch for this. But ok...

Issue #165, opened 2017-10-19
The TU Film announcements shouldn't appear in the news


TG908:
Since we are not going to refactor the news api we somehow need to make sure, that the tu film movies won't show up in the news card.

kordianbruck:
Why should they not appear in the news? 
The idea is to show one card for each news source. Thats how its done in Android.

TG908:
Then we should work on making one card for each news source. @mathiasquintero might have already done this in #167

mathiasquintero:
Sure! We can do it!

TG908:
duplicate see #220

PR #202, opened 2017-10-26
Release/1.3

### LINKED ISSUE #72: Poor performance due to missing caching
So from what I've tested the app does not seem to cache most of the data. Most noticebly it happend several times that the app was unresponsive due to load some data (10-20 seconds in the case of the calendar).

We will run into problems with the TUM. Especially the calendar uses a lot of resources on their mainframe. That should be updated once every 7 days or so at the very most! (Ask @xsrf for more details on this)

Please take a look at the Androids caching mechanics: https://github.com/TCA-Team/TumCampusApp/blob/master/app/src/main/java/de/tum/in/tumcampusapp/managers/CacheManager.java

### LINKED ISSUE #90: Refactor URLs
Refactor the creation of URLs by writing all the baseURLs in a JSON file . And finally get rid of all the warnings from Sonarqube

### LINKED ISSUE #109: Grades and Books should each be displayed in a single card
Todo: Create dedicated model for each of these cards

### LINKED ISSUE #118: More View crashes/doesnt display name

### LINKED ISSUE #121: User Details don't always match the exact user
See Ticket #880490

So far we get the user details by a search with the exact name and filter the list a bit afaik. Perhaps tinker with the filtering a bit or try to fetch it from the LRZ id.

### LINKED ISSUE #156: Caching Everything
If individual room maps images aren't too large, we should consider caching them on device.

Obvious advantage of this is performance: if we don't have to load the images from the internet, we can display them way faster if the user is on a slow connection.

### LINKED ISSUE #179: Remove TabBar

### LINKED ISSUE #190: Large Titles?

### LINKED ISSUE #194: Let's do another release on Sunday
Planned Features (would close 11 issues and 4 PRs !):

- [x] #192 Remove Tab Bar
- [x] #167 network refactor
- [ ] #119 add default campus (needs work)

Anything else?

### LINKED ISSUE #195: Open every url in a SFSafariController
For the sake of consistency we should open all urls in a safari controller and not in safari itself.

I'd very much like this in the 1.3 aswell

TG908:
Please test this.

Fixes #190
Fixes #179
Fixes #118

Fixes #72
Fixes #109 
Starts progress in #43 & #165
Fixes #156
Fixes #90
Fixes #195
Fixes #194
Fixes #121


mathiasquintero:
Since this branch has waaaay too much stuff here. I'd suggest we don't squash the commis this time in order to be able to separate them better in case something goes wrong.

And for the record in the future, we have to avoid MRs that span over more than one feature or bug...

TG908:
@mathiasquintero I think for future releases we should make PRs for the release branch and after release if everything goes well we merge the release branch into master

mathiasquintero:
@TG908 another idea would be to create a release branch. Always merge to master and cherry pick the features and bug fixes that will go to the release and merge those to the release branch

TG908:
The library card is still missing. It also doesn't show up when clicking on edit.

mathiasquintero:
@TG908 It is in edit. Under Book Rental

TG908:
The card doesn't work on any of my devices

mathiasquintero:
I understand. But I can't reproduce it without my account.

So you'll have to debug the functions where we fetch the rented books. I think we're failing to login

TG908:
It also seems like Editing the card order or removing cards doesn't work

mathiasquintero:
Fixed it

mathiasquintero:
@TG908 Going to more and back makes the navbar in the card view have a large title. Even without text.

mathiasquintero:
@TG908 is there anyway we can open the search results on the same navigation controller

TG908:
@mathiasquintero I think the search results controller is opened in a whole new ui window

mathiasquintero:
What if we keep a weak reference to the root vc and call push on those?

TG908:
> @TG908 Going to more and back makes the navbar in the card view have a large title. Even without text.

I cannot reproduce this

> What if we keep a weak reference to the root vc and call push on those?
I'll try

mathiasquintero:
Here you go! The text may not be there, but do you see the height difference?

simulator screen shot - iphone x - 2017-10-27 at 01 06 54
simulator screen shot - iphone x - 2017-10-27 at 01 06 58
simulator screen shot - iphone x - 2017-10-27 at 01 07 03

TG908:
@mathiasquintero is this the iphone x?

mathiasquintero:
Yes. But can be reproduced on any iphone

TG908:
> What if we keep a weak reference to the root vc and call push on those?
 
Would work will do a proper implementation tomorrow.

mathiasquintero:
@TG908 looks great now! I think we're done with most of what we wanted to do here. I'd say we are only missing the extra cards you wanted to make for like MVV, Library, Lectures, etc... 😉 

Perhaps you can get started on that in a separate branch starting from this one and we can later merge it back. So that we still follow the process

TG908:
@mathiasquintero I think the new cards and improved ui might take a while, because I'm quite busy. What I still need to do is take a look at the BookRental Card. Otherwise we are good. 😄

TG908:
Fixed Book Rentals we are ready for launch 🚀

TG908:
Ready for take of

mathiasquintero:
@mammuth have you checked this out?

mathiasquintero:
We also have to update the bundle version

mathiasquintero:
The build got stuck on `CalendarViewController.swift`. Dafuq?!?!

mathiasquintero:
Merging this now!

Issue #198, opened 2017-10-24
Add search bar to TUM.sexy redirects


mammuth:
As the TUM.sexy redirects are growing - add a search bar to the view which filters the list of redirects.

mathiasquintero:
I'd suggest showing them in the regular search as well

mathiasquintero:
Suggestion: Let TUMSexyManager implement SearchManager and filter links locally...

mathiasquintero:
This is my simples solution: https://github.com/mathiasquintero/Fuzzi

Which should allow us to search for both the domain and the description

PR #204, opened 2017-11-02
Implementing Tum Sexy Search

mathiasquintero:
### Issue
#198 

Search Tree is implemented as a BK-Tree over all the words mentioned in all the links mapping to an array of Links that include the words.

Implementation here:
https://github.com/mathiasquintero/Fuzzi/blob/master/Fuzzi/Classes/Searchable.swift#L20


mathiasquintero:
simulator screen shot - iphone x - 2017-11-02 at 11 32 55
This works by adding an extra category to the search result list

mathiasquintero:
The reason this was made so complicated was to allow for misspelling and fuzzy searches. Like this:
img_9221

Podfile
```diff
@@ -6,6 +6,7 @@ use_frameworks!
 
 target 'Campus' do
     pod 'Sweeft', '~> 0.13.3'
+    pod 'Fuzzi', '~> 0.1.0'
     pod 'MCSwipeTableViewCell', '~> 2.1.4'
```
TG908: This shouldn't be here anymore. I'll make a PR removing it

PR #207, opened 2017-11-07
Mathias/add memory cache

### LINKED ISSUE #193: Calender Freeze
"Leider gibt es immer noch Probleme mit dem Kalender - Wenn man zwischen Tagen hin und her wechselt bleibt die App oft eingefroren"

Feedback from a user. Are you parsing it on each swipe?

mathiasquintero:
Added an extra in-memory cache layer for better performance ;)

Fixes #193


TUM Campus App/Manager.swift
```diff
@@ -83,6 +83,33 @@ extension CachedManager {
...
+
+protocol MemoryCachedManager: class, CachedManager {
+    var cache: Cache<[DataType]>? { get set }
+    func performRequest(maxCache: CacheTime) -> Response<[DataType]>
+}
+
+extension MemoryCachedManager {
+    
+    func fetch(maxCache: CacheTime) -> Promise<[DataType], APIError> {
+        switch (maxCache, cache) {
```
mathiasquintero: @mammuth this is where the magic happens ;)

TUM Campus App/Manager.swift
```diff
@@ -83,6 +83,33 @@ extension CachedManager {
     
 }
 
+struct Cache<Value> {
```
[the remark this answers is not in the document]
mathiasquintero: Who the hell made that rule!?! Cache<V> is not always the best idea

mathiasquintero:
Do not merge yet. I found a bug...

TG908:
[this](https://github.com/Juanpe/SkeletonView) looks nice ... maybe we can indicate loading of data from web/cache or a missing connection

mathiasquintero:
I agree. I don't know if we want to indicate if it's cached but missing connection and still loading would be really nice.

I'll make an issue

TUM Campus App/CafeteriaManager.swift
```diff
@@ -54,7 +55,7 @@ final class CafeteriaManager: CachedManager, SingleItemCachedManager, CardManage
         }
     }
     
-    func fetch(maxCache: CacheTime) -> Response<[Cafeteria]> {
+    func performRequest(maxCache: CacheTime) -> Response<[Cafeteria]> {
```
lnsp: Why call it performRequest and not fetch? Every other *Manager implements fetch and the name change is a bit inconsistent. And it seems to do the same thing (correct me if I'm wrong).

TG908:
What is the difference between MemoryCachedManager and ChachedManager?

mathiasquintero:
The MemoryCachedManager is a CachedManager with an extra layer for cache.

The CachedManager caches at the request level.
The MemoryCachedManager will have an in memory version of the cache and after a miss go to the request based cache of the CachedManager. If that fails then it will perform the request

TUM Campus App/CafeteriaManager.swift
mathiasquintero: Yeah. I wanted to call it `fetch`, but then we are in the ambiguous naming territory because it will be called exactly the same.

We want to make sure that `fetch(maxCache)` still checks the in-memory cache first.
I'm open to ideas ;)

lnsp: Maybe `fetchFromSource` (since it reflects the relationship with `fetch` better) but I don't think that solves the ambiguity problem.

mathiasquintero: Maybe. The problem simply rises for using method composition to hide away abstractions. The second you have two default composited implementations the compiler will scream in agony for having names them identically.

A different approach would be having the fetch function performed by an private implementation of Cached Manager and using delegation. As in:

```swift
public class CafeteriaManager: MemoryCachedManager {

    private struct InternalManager: CachedManager { func fetch(maxCache: CacheTime) {} }

    // TODO: delegate fetching to internal manager on cache miss

}
```

TG908: too complicated.

mathiasquintero: What about wrapping it in a class instead of a protocol. Then the user will have more info on how caching works but I can live with that.: 

```swift
class MemoryCached<Manager: CachedManager>: CachedManager {



}
```

Issue #210, opened 2017-11-11
Implement tum newspread


TG908:
I think we should implement the slides from the [news screen](https://tum.newspread.de/page/newspreadLive) in the app. The api looks straight forward and easy to use: [rss](https://tum.newspread.de/page/newspreadLive/4.rss). Maybe using cards like [this](https://github.com/PaoloCuscela/Cards).

TG908:
![slice1](https://user-images.githubusercontent.com/7985149/32683942-40e4ecfa-c67f-11e7-95ab-0873e7758593.png)

TG908:
<img width="375" alt="image" src="https://user-images.githubusercontent.com/7985149/32684003-dcb7143c-c67f-11e7-802c-7fb583bc0c5e.png">

mathiasquintero:
Ewww: ```<content:encoded>&lt;img class='true reloader' height='480px' width='960px' src='https://storage.googleapis.com/tum-newspread-de/assets/e09d0063-af8c-463a-a54c-b69d829c1f91.jpeg' /&gt;</content:encoded>```

kordianbruck:
Guys. This is part of the News API from tumcabe. Please don't parse any other endpoints other than tumcabe and tumonline. Everything should be cached on our server to reduce load on other services and be able to fix parsing problems without releasing a new version!

mathiasquintero:
So I'd suggest we refactor our use of the news API first so that we can separate these by category.
See #43

TG908:
Categories:

- News (TUM News, Fachschaft News)
- TU Film (TU Film)
- newspread (newspread)

mathiasquintero:
@TG908 The Network part is implemented. You can use it but we're not receiving any of the newspread entries from the endpoint yet

mathiasquintero:
Turns out most items from the news spread are pretty old and don't come by default.

@kordianbruck I only now noticed that Android uses the last id in sync. 
Should we do that? If so do we really have to store every news entry on the device?

kordianbruck:
@mathiasquintero yes, that would be a great idea to reduce the amount of traffic and server load. Its really easy to implement. 
I wouldn't delete the entries. I don't think they delete the last id. So in case nothing new is there, they won't fetch anything. Newsspread is unfortunately broken right now. I'll have to check the cronjob for that as well.

mathiasquintero:
@kordianbruck Sounds good.
I could recreate that behavior with this:

```swift
[...]

PR #214, opened 2017-11-13
Split NewsManager by Sources

mathiasquintero:
### Issue
Fixes: #43 
Fixes: #210 (At least the API part)

### DIFF
--- a/TUM Campus App/News.swift
+++ b/TUM Campus App/News.swift
@@ -11,14 +11,21 @@ import Sweeft
 
 final class News: DataElement {
     
+    enum Source: String {
+        case movie = "2"
+        case regular = "1"
+    }
+    
     let id: String
+    let source: Source
     let date: Date
     let title: String
     let link: String
     let image: Image
     
-    init(id: String, date: Date, title: String, link: String, image: String? = nil) {
+    init(id: String, source: Source, date: Date, title: String, link: String, image: String? = nil) {
         self.id = id
+        self.source = source
         self.date = date
         self.title = title
         self.link = link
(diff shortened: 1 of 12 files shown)


TUM Campus App/News.swift
```diff
@@ -11,14 +11,21 @@ import Sweeft
 
 final class News: DataElement {
     
+    enum Source: String {
+        case movie = "2"
+        case regular = "1"
```
kordianbruck: Source is a primary key in the table of the newsSources. Its currently right, that the ID 2 is the TU Film, but the ID can be anything. You should account for that as we add new sources in the future.

Here are all sources: 
screenshot

mathiasquintero: Ok. So do we expect the source ids to change over time?

TUM Campus App/NewsSource.swift
```diff
@@ -0,0 +1,78 @@
...
+    static let operatingSystems = News.Source(identifier: 6)
+    static let newsspreadFMI = News.Source(identifier: 7)
+    static let newsspreadChemistry = News.Source(identifier: 8)
+    static let newsspreadSG = News.Source(identifier: 9)
+    static let newsspreadWI = News.Source(identifier: 10)
+    static let newsspreadPhysics = News.Source(identifier: 11)
+    static let newsspreadLab = News.Source(identifier: 12)
+    static let newsspreadWeihenstephan = News.Source(identifier: 13)
+    static let alumni = News.Source(identifier: 14)
+    static let impulsiv = News.Source(identifier: 15)
```
kordianbruck: Thats more like it :heart:

mathiasquintero: I also love bitmasks

TUM Campus App/NewsSource.swift
```diff
@@ -0,0 +1,78 @@
...
+        .newsspreadChemistry,
+        .newsspreadSG,
+        .newsspreadWI,
+        .newsspreadPhysics,
+        .newsspreadLab,
+        .newsspreadWeihenstephan,
+        ]
+    
+    static var restOfNews: News.Source {
+        return News.Source.all.subtracting([.movies, .newsSpread])
```
mathiasquintero: I hope you're cool with this. Everything that is not in movies or any of the newsSpred sources falls here

kordianbruck: Yea, thats probably the best way to handle it.

PR #219, opened 2017-11-14
Location Manager Fix

### LINKED ISSUE #218: Fix Location Manager
```
2017-11-14 19:05:36.431704+0100 Campus[14956:6882878]
A location manager (0x1021ba7f0) was created on a dispatch queue executing on a thread other 
than the main thread.  It is the developer's
responsibility to ensure that there is a run loop running on the 
thread on which the location manager object is allocated. 
In particular, creating location managers in arbitrary dispatch queues (not attached
to the main queue) is not supported and will result in callbacks not being received.

```

Should we investigate this?

TG908:
Fixes #218

### DIFF
--- a/TUM Campus App/Extensions.swift
+++ b/TUM Campus App/Extensions.swift
@@ -11,7 +11,8 @@ import UIKit
 import SafariServices
 import Sweeft
 
-private let locationManager = CLLocationManager()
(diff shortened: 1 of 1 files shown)


TG908:
It's gone

mathiasquintero:
Ok. Will merge when CI is done

PR #226, opened 2017-11-17
Location Manager Fix again

TG908:
Some further testing resulted into deadlocks. Caused by creating the Location Manager from the main thread. So I am catching these possible deadlocks before they occur.


TG908:
This should be merged

PR #246, opened 2018-04-09
Tim/pod update+objc protocol fix

TG908:
I've fixed a number of warnings, updated function signatures (objc compatibility), updated some pods, updated the xcode project settings, enabled wmo and some pods are now compiled for swift 4 while others are compiled for swift 3.3

This should work with Swift 4.0.3 (Xcode 9.2) [download](https://swift.org/builds/swift-4.0.3-release/xcode/swift-4.0.3-RELEASE/swift-4.0.3-RELEASE-osx.pkg).

PR #249, opened 2018-05-01
Login Logo was not properly sized

TG908:
Fixed logo view on login screen to look like the one in the CardViewController

**Before**
simulator screen shot - iphone 8 plus - 2018-05-01 at 12 50 04
**After**
simulator screen shot - iphone 8 plus - 2018-05-01 at 12 51 47


TUM Campus App/LoginViewController.swift
```diff
@@ -91,6 +92,17 @@ class LoginViewController: UIViewController {
...
     }
+    
+    private func setupLogo() {
+        let bundle = Bundle.main
+        let nib = bundle.loadNibNamed("TUMLogoView", owner: nil, options: nil)?.flatMap { $0 as? TUMLogoView }
+        guard let view = nib?.first else { return }
+        logoView = view
+        view.frame = CGRect(x: 0, y: 0, width: 100, height: 40)
+        view.widthAnchor.constraint(equalToConstant: 100).isActive = true
+        view.heightAnchor.constraint(equalToConstant: 40).isActive = true
```
kordianbruck: I'm not sure about how this works on iOS, but on Android we have many different resolutions of different phones. Does this automatically scale down&up? From the removed code below, I would guess that would've been the intention. (Wouldn't it be great to have comments to see that the original author thought? :P )

thellmund:
@TG908 I can’t get this to build

TG908:
@thellmund This PR requires #246 to be merged... In #251 these changes are already implemented.

TUM Campus App/LoginViewController.swift
kordianbruck: @thellmund any comments on this?

thellmund: @kordianbruck As far as I know, iOS handles these things like so: The Navigation Bar is _always_ 44 pixels high. On high-res devices, it’s simply drawn at 2x (88 device pixels), and on iPhone X at 3x (132 device pixels). Thus, @TG908’s implementation looks good to me.

PR #258, opened 2018-06-15
Fix RefreshControl glitch in NavigationBar

### LINKED ISSUE #255: Wrong height of navigation bar
When opening the app, the navigation bar is displayed with a wrong height. It stays like this until you scroll the CollectionView.

simulator screen shot - iphone 7 plus - 2018-06-14 at 14 17 12

thellmund:
This commit fixes #255 (wrong height of the NavigationBar).


thellmund:
It appeared to me as if it was a problem with the standard component, which is why I moved the refresh control to the top of the collection view. I'll have a look at your commit later.

PR #260, opened 2018-06-16
Implement card ordering

thellmund:
This commit implement card ordering. Cards in `CardViewController`are now ordered as specified in `EditCardsViewController`.

### DIFF
--- a/TUM Campus App/BookRentalManager.swift
+++ b/TUM Campus App/BookRentalManager.swift
@@ -10,7 +10,8 @@ import Foundation
 import Sweeft
 import Kanna
 
-final class BookRentalManager: SingleItemManager, CardManager {
+@available(*, deprecated)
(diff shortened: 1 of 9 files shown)


TG908:
I liked the old code better ... `var cardKeys: [CardKey] { return PersistentCardOrder.value.cards }`

thellmund:
I'll have a look at the sorting extension 👍 

As this is work in progress, I planned on adding some caching mechanism for sortedDataSources before submitting it for review

thellmund:
@TG908 I’ve just updated the branch with the sorting extension that you posted above as well as with caching. Let me know how you think it can be improved.

TG908:
It would be simple to have sortedDataSources and allDataSources as instance variables, so you can get rid of the duplicate variable and the optional. You can then just set the filtered and sorted dataSources in init() and refresh()

PR #261, opened 2018-06-16
Add card content interaction

thellmund:
These changes make card content interactable, so that we perform an action when the user selects the content of a card.

### DIFF
--- a/TUM Campus App.xcodeproj/project.pbxproj
+++ b/TUM Campus App.xcodeproj/project.pbxproj
@@ -204,6 +204,7 @@
 		9D10C2AB20D289AA00DB41EE /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 9D10C2AA20D289AA00DB41EE /* GoogleService-Info.plist */; };
 		9D77671120D033F800F9A4FD /* UITextViewNoStartInset.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9D77671020D033F800F9A4FD /* UITextViewNoStartInset.swift */; };
 		9D77671720D0494D00F9A4FD /* GradeColor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9D77671620D0494D00F9A4FD /* GradeColor.swift */; };
+		9DCEDE5D20D461D300896635 /* RoundBlurredBackground.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9DCEDE5C20D461D300896635 /* RoundBlurredBackground.swift */; };
 		B20DF5B992DFD30580046509 /* Pods_TUM_Campus_AppUnitTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 63ABF4E4DF444D3822681E59 /* Pods_TUM_Campus_AppUnitTests.framework */; };
 		E1369C9D1F87D25E00D6766E /* ViewControllerProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1369C9C1F87D25E00D6766E /* ViewControllerProvider.swift */; };
 /* End PBXBuildFile section */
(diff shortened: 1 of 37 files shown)


thellmund:
Until now, I’ve implement interaction for news items, newspread items, film items and for the tuition fees card. 

@kordianbruck @TG908 Open questions:
- What is the desired behavior for the rest? E.g. when tapping within the MVV stations card, what should hapen?
- Should we limit the number of items displayed and add a “Show all” button in the top right of every card? Particularly for cafeterias and MVV stations, I think this makes sense.

thellmund:
Let’s align on the design of the calendar cells on Tuesday. 

The problem with MovieDetailsTableViewController is that it expects a `Movie` whereas we currently use `News` objects to represent movies. We can change this in a separate pull request and then show the details view instead of the web view.

kordianbruck:
> What is the desired behavior for the rest? E.g. when tapping within the MVV stations card, what should hapen?

We also have a MVV Activity with a longer list of departures. So show like the next 4 in the card and the next 10 in the full view.

> Should we limit the number of items displayed and add a “Show all” button in the top right of every card? Particularly for cafeterias and MVV stations, I think this makes sense.

I don't think this flow works for everything, but we might have to think about limiting what exactly we show in the home screen. Overwhelming the user with too much information may be just as bad.

thellmund:
@kordianbruck I’ve implemented a first version of this. Any feedback? Does it make sense? (also @TG908, of course 😊)



Some insight into the changes: 
- Each horizontal CollectionView now only displays a limited number of items. For single items, the limit is 10. For composite items (e.g. MVG station, which contains departures), the limit is 5. 
- When tapping on “Show more”, we now pass the already downloaded information to the ViewController. 

Open issue: 
- What should we do for individual MVG stations? My idea is to include a “More” button as the last item in the list of departures, which the user can tap to jump directly to the `DeparturesViewController`.

TG908:
> Open issue:
> 
> What should we do for individual MVG stations? My idea is to include a “More” button as the last item in the list of departures, which the user can tap to jump directly to the DeparturesViewController.

Sounds good

TG908:
I would prefer something more like this:
simulator screen shot - iphone 8 plus - 2018-06-21 at 01 07 56

thellmund:
@TG908 That looks nice! Feel free to push your changes to this branch 😊

thellmund:
Sorry, I intended to add the rationale in the commit message 🙈 

The shadow does not really work for the movie posters for two reasons: 
1. In the other cards, we use the shadow to elevate the entire CollectionViewCells. However, in the TUFilmCollectionViewCell, we only had the shadow behind the image view rather than behind the entire cell, which is inconsistent. 
2. It interferred with the label below the movie poster and resulted in visual disturbance.

In the other horizontal collection views, we use shadows behind the complete item, not behind an individual element. For instance in the news collection view cell, we do not have a shadow behind the image, but only behind the entire item.

TG908:
> Sorry, I intended to add the rationale in the commit message 🙈 
> 
> The shadow does not really work for the movie posters for two reasons: 
> 1. In the other cards, we use the shadow to elevate the entire CollectionViewCells. However, in the TUFilmCollectionViewCell, we only had the shadow behind the image view rather than behind the entire cell, which is inconsistent. 

I agree on the inconsistency, but I think it looks slightly nicer this way.

> 2. It interferred with the label below the movie poster and resulted in visual disturbance.

I couldn't reproduce any interference in my testing.

> In the other horizontal collection views, we use shadows behind the complete item, not behind an individual element. For instance in the news collection view cell, we do not have a shadow behind the image, but only behind the entire item. 


Also during my testing I noticed that the calendar view has a few shortcomings:

- The most important information (the event title) has the smallest font
- The whole screen is cluttered
- Some key info should be visually highlighted by using a different color or font size for example
- The layout doesn't work on smaller screens like the iphone se.

thellmund:
Maybe then this is the right time to discuss if we actually need the movie title. It’s always in the poster, isn’t it? If we decided to remove the label, I’m alright with adding a shadow to the movie poster. 

Calendar: Other than adding the background and reducing the big spacing between the title label and the location label, I did not change anything about the calendar cards (at least consciously). The small font size occurs because the system is squeezing the text into the label, reducing its font size. We can think about deactivating this, though that may make the cards less informative. 

Can you post screenshots with the other points about the calendar cards that you’ve mentioned?

I’ll have a look at the odd cell sizes.

thellmund:
Regarding the odd cell sizes: It’s due to the current use of `preferredHeight`, a property of each `TUMDataSource` that sets the cell height. Instead of this static property, each `TUMDataSource`should provide its own size via a protocol method. I’ll add an issue that we should address in a separate pull request.

TG908:
> Maybe then this is the right time to discuss if we actually need the movie title label. It’s always in the poster, isn’t it? If we decided to remove the label, I’m alright with adding a shadow to the movie poster.

Ok lets get rid of the titles.

> Calendar: Other than adding the background and reducing the big spacing between the title label and the location label, I did not change anything about the calendar cards (at least consciously). The small font size occurs because the system is squeezing the text into the label, reducing its font size. We can think about deactivating this, though that may make the cards less informative.
> 
> Can you post screenshots with the other points about the calendar cards that you’ve mentioned?
> 
> I’ll have a look at the odd cell sizes.


simulator screen shot - iphone se - 2018-06-24 at 22 34 51
simulator screen shot - iphone se - 2018-06-24 at 22 35 19

thellmund:
I removed the red color of the countdown because it’s totally unnecessary. I increased the font size because it was way too small to be considered legible. 

Regarding the number of lines in the title: If we have horizontally scrolling cards, we ideally want to have cards that have a uniform height. Otherwise, I find that it looks somewhat odd.

PR #263, opened 2018-06-16
Fixed NavBar Glitch

### LINKED ISSUE #255: Wrong height of navigation bar

TG908:
This fixes #258 and fixes #255


thellmund:
It this ready to be reviewed?

TG908:
Yes

thellmund:
The current implementation still has an odd behavior. This is what the nav bar looks like when launching the app, without any user interaction. 

simulator screen shot - iphone 7 - 2018-06-16 at 21 32 01

TG908:
I patched it by just not displaying the loading animation on app launch (which btw no other app does so its fine).

thellmund:
I don’t think we can make this comparison that easily. When launching, most apps get cached data from their on-device database and display it immediately. Simultaneously, they fetch the newest content from a remote source and display a network/loading indicator in the status bar. However, we currently don’t store data locally and always must fetch it from a remote source. Depending on the connection, this can take a while. 

Not showing any loading indicator might give the user the impression that the app is stuck. For that reason, I’m advocating for adding an indicator of some sort. If the refresh indicator causes glitches, we can think about adding a loading indicator in the center of the screen.

thellmund:
Looking good to me now 👍

PR #264, opened 2018-06-16
Improve design of onboarding screens

thellmund:
The current onboarding screens have the following problems: 
- The text is a mix of German and English. 
- The text-only layout is not very inviting. 
- The spelling of TUMonline is wrong. 

Here’s a potential update to the onboarding screens: 
onboarding

Let me know what you think. 

Open questions: 
- The three separated textfields are odd. For a cleaner layout, they could be replaced with a button that opens an alert with a textfield, where the user can enter their TUM ID. Opinions?
- We could use some sort of illustration as eyecandy, for instance the sketch of the TUM tower. Opinions?


TG908:
> Let me know what you think.
I like the screenshots.
> Open questions:
> 
> The three separated textfields are odd. For a cleaner layout, they could be replaced with a button that opens an alert with a textfield, where the user can enter their TUM ID. Opinions?
I actually like the current layout. It matches the structure of a TUM ID.

> We could use some sort of illustration as eyecandy, for instance the sketch of the TUM tower. Opinions?
I don't know about the TUM tower. But from user feedback we know that the token activation process could use some illustration.

kordianbruck:
@thellmund nice! Could we also in one go make `connecting to TUMonline` optional, as it is in the Android version?

TG908:
It already is

kordianbruck:
@TG908 cool beans - also with the same naming? "Connect to..."

TG908:
Its right under the login button. It’ says continue without tum Id

TUM Campus App/LoginViewController.swift
```diff
@@ -31,81 +30,71 @@ class LoginViewController: UIViewController {
...
-        confirmButton.alpha = textFieldContentsAreValid() ? 1 : 0.5
+    @IBAction func openInputDialog() {
+        let alert = UIAlertController(title: "Login", message: "Enter your TUM ID", preferredStyle: .alert)
+        alert.addTextField(configurationHandler: nil) // TODO
+        alert.addAction(
+            UIAlertAction(title: "Cancel", style: .cancel, handler: nil)
+        )
+        alert.addAction(
+            UIAlertAction(title: "Continue", style: .default) { action in
+                guard let text = alert.textFields![0].text else {
```
TG908: Do you need to force unwrap here?

thellmund: Good catch 👍 I changed it to a chain of optionals.

thellmund:
@TG908 This branch is still a work in progress, as I’ll have to adjust the logic to accommodate the App Store reviewers, which don’t have a TUM ID. You can revoke your approval again 😄

TG908:
@thellmund I think they do have one.

thellmund:
@TG908 Technically yes, but it does not conform to the usual pattern (ga12abc) and is not 7 characters long.

TG908:
I don't really like the new login prompt I liked the old one better. Maybe we can add a storyboard transition to the old screen?. For the appstore review problem, you could, check each of the 3 fields if it matches the letters of the testing account login. If all fields are used up and all 3 match the test login username display a prompt or add another text field at the end

kordianbruck:
@TG908 I actually like the Popup - it makes the signup process to be super simple and cleaner. Can we bring this branch to a mergeable state?

TG908:
closed due to inactivity

kordianbruck:
@TG908 till will probably not contribute too much to iOS in the future. I guess you don't want to change the onboarding?

TG908:
@kordianbruck onboarding is not my first priority.

kordianbruck:
@TG908 alright, then just lets keep the branch around, in case somebody wants to pick this up in the future.

PR #265, opened 2018-06-17
Implemented error view and network timeout

TG908:
We now display an error message after trying for 10 seconds to load the cards.

simulator screen shot - iphone 8 plus - 2018-06-17 at 03 12 30

Some of the logic depends on #260

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 [![Travis](https://api.travis-ci.org/TCA-Team/iOS.svg?branch=master)]()
 
-# TumCampusApp - an unoffical guide through university life
+# TumCampusApp - an unofficial guide through university life
 
 The [TUM Campus App (TCA)](https://itunes.apple.com/app/id1217412716) is an open source project, developed by volunteers and available on the App Store.
 
 The TCA mostly targets phones, but can also be used on tablets or any other device that runs iOS. This is the repo for the iOS Version of the TUM Campus App.
 
 ## Work In Progress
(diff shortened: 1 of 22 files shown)


thellmund:
I think the placeholder text should be less technical, along the lines of “Something went wrong”. Additionally, it should provide the user with an option to try again.

TG908:
Updated version:
simulator screen shot - iphone 8 plus - 2018-06-21 at 01 07 56

thellmund:
I like the design. Two questions: 
1. The header text size looks a little too big. Could you try something like Title 2?
2. What’s the rationale behind using a stack view with empty views to center the try again button? Can you try achieving the same design without them?

thellmund:
I like the design. Two questions:
1. The header text size looks a little too big. Could you try something like Title 2?
2. What’s the rationale behind using a stack view with empty views to center the try again button? Can you try achieving the same design without them?

TG908:
I think the font size matches the one used by apple in the App Store quite well

img_2401

TG908:
> What’s the rationale behind using a stack view with empty views to center the try again button? Can you try achieving the same design without them?

Its the easiest way to center the button. It's unnecessarily complicated to archive the same thing in another way.

thellmund:
> Its the easiest way to center the button.
Centering it horizontally via a constraint should be a little cleaner, I think.

thellmund:
> Centering it horizontally via a constraint should be a little cleaner, I think.

Oh, it within a StackView. Then it’s fine 👌

PR #273, opened 2018-06-20
Fix film poster size

thellmund:
This change sets a fixed TUFilm poster size, independent of the number of lines in the title label.


thellmund:
@TG908 There’s still an issue: The first line of the label has a tiny indent. I was not able to figure out how. Any idea of why that might be?

thellmund:
Fixed it. Turns out we were not properly trimming the title string, so they always had a leading space.

PR #280, opened 2018-06-30
Fix wrong cell heights

thellmund:
This pull request fixes the wrong height that some CollectionViewCells experienced. It achieves this by introducing: 
1. `FixedColumnsFlowLayout`: A layout that always shows a fixed number of columns (iPhone: 1, iPad: 2 in portrait, 3 in landscape)
2. `VariableColumnsFlowLayout`: A layout sets the width of items based on a provided aspect ratio. Used for movie posters and newspread news.


TUM Campus App/CollectionViewLayouts/ColumnsFlowLayout.swift
```diff
@@ -0,0 +1,54 @@
...
+//  ColumnsFlowLayout.swift
+//  Campus
+//
+//  Created by Till on 30.06.18.
+//  Copyright © 2018 LS1 TUM. All rights reserved.
+//
+
+import UIKit
+
+class ColumnsFlowLayout: NSObject, UICollectionViewDelegateFlowLayout {
```
TG908: The naming isn't accurate, since this is not a FlowLayout but rather a FlowLayoutDelegate

TUM Campus App/CollectionViewLayouts/ColumnsFlowLayout.swift
```diff
@@ -0,0 +1,54 @@
...
+    
+    var isPhone: Bool {
+        return UIDevice.current.userInterfaceIdiom == .phone
+    }
+    
+    var isLandscape: Bool {
+        return UIDevice.current.orientation.isLandscape
+    }
+    
+    var columns: CGFloat {
```
TG908: Using a CGFloat here does't seem right. Sure it makes further computations easier but...

thellmund: You’re right. Changed it.

TUM Campus App/CollectionViewLayouts/ColumnsFlowLayout.swift
thellmund: Alright, renamed these classes.

thellmund:
@TG908 Implemented the requested changes. Could you have a look again?

PR #303, opened 2018-07-19
Allow toggling between the default logo and the pride version

lukaskollmer:
This also:
- renames `tumLogoTap(sender:)` to `toggleLogo` and removes the unused `sender` argument
- removes the `@IBAction` attribute in favor of `@objc` (allowing us to make the method private)

tbh i'm not really sold on the logo switching code but i guess that's nicer than introducing some additional property to keep track of which version of the logo is currently being displayed

### DIFF
--- a/TUM Campus App/TUMLogoView.swift
+++ b/TUM Campus App/TUMLogoView.swift
@@ -22,15 +22,17 @@ import UIKit
 
 
 class TUMLogoView: UIView {
+    private static let defaultLogo = #imageLiteral(resourceName: "logo-blue")
+    private static let rainbowLogo = #imageLiteral(resourceName: "logo-rainbow")
     
     @IBOutlet weak var imageView: UIImageView!
     
-    @IBAction func tumLogoTap(sender: AnyObject) {
-        imageView.image = #imageLiteral(resourceName: "logo-rainbow")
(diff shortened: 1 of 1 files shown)


thellmund:
Code is looking good, but a few other questions: 
1. The choice is not persisted. Is that the desired behavior? If so, why?
2. What about the logo on the launch screen? In the Android version, we use this logo to allow the user to toggle the default logo and the pride logo throughout the entire app. Should the iOS version adopt the same behavior? (what do you think, @kordianbruck and @TG908?)

thellmund:
@lukaskollmer I did not realize that an option to switch to the pride logo already exists. In this case, the change looks good to me.

TG908:
There is an API to change the app icon...

lukaskollmer:
yeah you're right i thought @thellmund was talking about the launch screen. 

we can definitely change the spring board icon as well, however that should be a separate PR, since we'd need to replace the currently used asset catalogs w/ manually specified filenames in the info.plist, which can be bit of a pita to get working.

also, if we allow changing the spring board icon, we should probably also add a "blue logo on white background" option.

---
Lukas Kollmer
lukas.kollmer@gmail.com

kordianbruck:
@thellmund as for your question: yes, persistency  would be nice

Issue #297, opened 2018-07-10
Layout for MVG Station View needs improvements


TG908:
The image says it all:

![img_27cee28b2c62-1](https://user-images.githubusercontent.com/7985149/42527154-2c5177ec-8478-11e8-8f0d-ab52b19d3ed0.jpeg)

ffittschen:
In the AppStore version of the app it is still an issue, can I work on this as part of Hacktoberfest?

lukaskollmer:
@ffittschen sure!

TG908:
I was going to add a timer to the MVG card if you were planning to do the same let me know.

ffittschen:
@TG908 Did you already start? I reused the style of the Card, so I could also extract it into one component as well. This would result in both having the timer and looking like the departure in the Card (see screenshot)

<img width="320" alt="screen shot 2018-10-16 at 13 53 06" src="https://user-images.githubusercontent.com/7734806/47014714-585c5a80-d14b-11e8-8b1c-b5c827494e83.png">

TG908:
I didn’t start yet. If you could make this work that‘ll be great.

PR #311, opened 2018-10-17
Fix MVG station layout

ffittschen:
### Issue

#297 

As discussed in the issue, I extracted the layout of the individual departures from the card into an own `DepartureView` that includes the countdown timer and is reused for the card and the detailed table view.

### Screenshots

The look of the MVG Stations `CardViewController` didn't change.
New look of the `DeparturesViewController`:
screen shot 2018-10-16 at 13 53 06


kordianbruck:
Cheers!

Issue #95, opened 2017-04-17
Localisation


TG908:
At this point we should think about localizing our app because it is a mix of german and english at the moment. We should at least feature german and english. Maybe we can get the TUM Language Cafe involved, and translate the app in some more languages. We could just print some templates and leave them at the tables to be translated.

How is localization handled in the android app?

kordianbruck:
Android: natively from the SDK supports translation of all strings via an .XML file. (Strings.xml)

TG908:
Do we have any translated strings files?

kordianbruck:
Yes, you just need to check the different Values folders: https://github.com/TCA-Team/TumCampusApp/tree/master/app/src/main/res
(values-de/ru/fr/es/it)

TG908:
no longer applicable due to app rewrite

PR #327, opened 2019-02-15
Rewrite Everything

TG908:
I am completely rewriting the app from ground up 😅

PR #328, opened 2019-05-03
add localization for German language

vollkorntomate:
### Issue
The app was originally only available in English. I added support for German language. Other languages may be added easily.

Localisation #95

### Screenshots
No screenshots.


kordianbruck:
@vollkorntomate thanks for contributing. Currently we are signifincantly reworking the app in #327 - @TG908 can we still merge this up front and will it be useful?

TG908:
I think we can reuse the .string files at some point. Merging this now won't be useful.

TG908:
no longer applicable due to app rewrite

PR #335, opened 2020-06-09
[RoomFinder] show error message, add missing localisations

lukaskollmer:
- Show a label with an error message if a room couldn't be found. This seems better than just displaying an empty table
- Added an explicit check if a room finder API request was cancelled, since for some reason the completion handler would still get invoked even after calling `sessionManager.cancelAllRequests()`
- Added some missing localisations to the german string file

### DIFF
--- a/TUM Campus App/Room Finder/RoomFinderViewController.swift
+++ b/TUM Campus App/Room Finder/RoomFinderViewController.swift
@@ -10,7 +10,6 @@ import UIKit
 import Alamofire
 
 final class RoomFinderViewController: UITableViewController, UISearchResultsUpdating {
-    private let errorMessageLabel = UILabel()
     private let sessionManager: Session = Session.defaultSession
     private var dataSource: UITableViewDiffableDataSource<Section, Room>?
 
@@ -29,11 +28,6 @@ final class RoomFinderViewController: UITableViewController, UISearchResultsUpda
         navigationItem.searchController = searchController
         definesPresentationContext = true
         tableView.tableFooterView = UIView()
-        errorMessageLabel.textAlignment = .center
-        errorMessageLabel.textColor = .secondaryLabel
-        errorMessageLabel.font = .systemFont(ofSize: 19)
-        errorMessageLabel.numberOfLines = 0
-        
         setupDataSource()
     }
 
@@ -46,17 +40,6 @@ final class RoomFinderViewController: UITableViewController, UISearchResultsUpda
 
             return cell
         }
-    }
-    
-    /// Shows an error message stating that we were unable to find the room.
-    /// - parameter roomName: The name of the room we were searching for. `nil` to remove the error message
-    private func showUnableToFindRoomErrorMessage(roomName: String?) {
-        guard let roomName = roomName else {
-            tableView.backgroundView = nil
-            return
-        }
-        tableView.backgroundView = errorMessageLabel
-        errorMessageLabel.text = NSString(format: "Unable to find room".localized as NSString, roomName) as String
     }
     
 
@@ -81,7 +64,7 @@ final class RoomFinderViewController: UITableViewController, UISearchResultsUpda
             // Cancel currently running requests and clear the table when searching for an empty string.
             // This seems preferable over sending the empty string to the API and clearing the table via the (expectedly) empty response
             sessionManager.cancelAllRequests()
-            showUnableToFindRoomErrorMessage(roomName: nil)
+            self.removeBackgroundLabel()
             return
         }
         let endpoint = TUMCabeAPI.roomSearch(query: searchString)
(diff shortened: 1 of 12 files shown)


TUM Campus App/Room Finder/RoomFinderViewController.swift
```diff
@@ -10,6 +10,7 @@ import UIKit
 import Alamofire
 
 final class RoomFinderViewController: UITableViewController, UISearchResultsUpdating {
+    private let errorMessageLabel = UILabel()
```
TG908, line 13: We have an extension for `UITableViewController` and `UIcollectionViewController` `setBackgroundLabel(with text: String)` to set the background label for error messages.

TUM Campus App/Room Finder/RoomFinderViewController.swift
```diff
@@ -41,6 +47,18 @@ final class RoomFinderViewController: UITableViewController, UISearchResultsUpda
...
+    
+    /// Shows an error message stating that we were unable to find the room.
+    /// - parameter roomName: The name of the room we were searching for. `nil` to remove the error message
+    private func showUnableToFindRoomErrorMessage(roomName: String?) {
+        guard let roomName = roomName else {
+            tableView.backgroundView = nil
+            return
+        }
+        tableView.backgroundView = errorMessageLabel
+        errorMessageLabel.text = NSString(format: "Unable to find room".localized as NSString, roomName) as String
```
TG908, line 59: Wouldn't it be nice to do something like that:  `errorMessageLabel.text = "\("Unable to find room".localized) roomName"` instead of using `NSString`?

TUM Campus App/Room Finder/RoomFinderViewController.swift
```diff
@@ -52,16 +70,35 @@ final class RoomFinderViewController: UITableViewController, UISearchResultsUpda
...
-        guard let searchString = searchBar.text else { return }
+        guard let searchString = searchController.searchBar.text else {
+            return
+        }
+        guard !searchString.isEmpty else {
+            // Cancel currently running requests and clear the table when searching for an empty string.
+            // This seems preferable over sending the empty string to the API and clearing the table via the (expectedly) empty response
+            sessionManager.cancelAllRequests()
+            showUnableToFindRoomErrorMessage(roomName: nil)
+            return
```
TG908, line 85: 👍🏼

TUM Campus App/Room Finder/RoomFinderViewController.swift
```diff
@@ -52,16 +70,35 @@ final class RoomFinderViewController: UITableViewController, UISearchResultsUpda
...
-        sessionManager.request(endpoint).responseDecodable(of: [Room].self, decoder: JSONDecoder()) { [weak self] response in
+        let request = sessionManager.request(endpoint)
+        request.responseDecodable(of: [Room].self, decoder: JSONDecoder()) { [weak self] response in
+            guard !request.isCancelled else {
+                // cancelAllRequests doesn't seem to cancel all requests, so better check for this explicitly
+                return
+            }
             var snapshot = NSDiffableDataSourceSnapshot<Section, Room>()
             snapshot.appendSections([.main])
+            if (response.value ?? []) == [] {
```
TG908, line 97: this looks cleaner:

```
let value = response.value ?? []

var snapshot = NSDiffableDataSourceSnapshot<Section, Room>()
snapshot.appendSections([.main])
snapshot.appendItems(value, toSection: .main)
self?.dataSource?.apply(snapshot, animatingDifferences: true)

if value == [] {
         self?.setBackgroundLabel(with: "\("Unable to find room".localized) \(searchString)")
} else {
        self?.tableView.backgroundView = nil
}
```

TUM Campus App/Room Finder/RoomFinderViewController.swift
lukaskollmer: because the substitution happens at different places within the string:
- "Unable to find room 'xx.xx.xxx'"
- "Raum 'xx.xxx.xx' konnte nicht gefunden werden".

There's no german phrasing I can think of where it'd sound natural having the room number at the end of the sentence.

Also, if there's additional languages being added, these translations too would have to be phrased in a way where the room number is at the end.

lukaskollmer: Also using `NSString` ?should? effectively be a noop, since it'll share the storage from the `String`, this is really just about getting the formatting api

TUM Campus App/Room Finder/RoomFinderViewController.swift
lukaskollmer: yeah you're right

TUM Campus App/Room Finder/RoomFinderViewController.swift
TG908: Cool. I didn't think of that. You are right

lukaskollmer:
@TG908 I implemented your feedback (using the background label thing and fixing the `?[]) ==[]` thing). While i was at it i noticed that the background label api was kinda inconsistent (there's `setBackgroundLabel(with:)` for setting a background label, but no corresponding method for removing it, and there was the exact same implementation twice (for `UITableViewController` and `UICollectionView`.

I attempted to improve that, by:
- adding a method to remove the backgruond label (i'd say `removeBackgroundLabel()` is more explicit than `tableView.backgroundView = nil`)
- putting the actual background label creation/removal code into a protocol, which `UITableView{Controller}` and `UICollectionView{Controller}` conform to,
- thus now consistently offering same the background-label functionality for both table and collection views, as well as their -controller counterparts

Yes this is a bit more convoluted than it was before, at least at the implementation side, but I think that using the api should be clearer/easier now

PR #336, opened 2020-08-10
consistently name cafeterias tab bar item "Cafeterias"

lukaskollmer:
Problem: when launching the app, the 3rd tab bar item is labeled "Cafeteria", and selecting it causes the name to change to "Cafeterias".

Fix: always label it "Cafeterias"


TG908:
Thank you!

PR #339, opened 2020-08-29
Added search bar to tum.sexy

christianschmucker:
I added a UISearchController for searching specific links in tum sexy. The filtered TUMSexyLinks are stored in filteredData.

### DIFF
--- a/TUM Campus App/TUMSexy/TUMSexyTableViewController.swift
+++ b/TUM Campus App/TUMSexy/TUMSexyTableViewController.swift
@@ -9,18 +9,27 @@
 import UIKit
 import CoreData
 
-final class TUMSexyTableViewController: UITableViewController {
+final class TUMSexyTableViewController: UITableViewController, UISearchResultsUpdating {
+    
     typealias ImporterType = Importer<TUMSexyLink,[String: TUMSexyLink],JSONDecoder>
     
     private static let sortDescriptor = NSSortDescriptor(keyPath: \TUMSexyLink.linkDescription , ascending: false)
     private let importer = ImporterType(endpoint: TUMSexyAPI(), sortDescriptor: sortDescriptor)
-
+    
+    let searchController = UISearchController(searchResultsController: nil)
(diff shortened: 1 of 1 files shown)


nerdsupremacist:
We used to have tum sexy search before. I guess the last refactor removed it somehow. And in fact it was implemented as fuzzy search using this FW I built back then:

https://github.com/nerdsupremacist/Fuzzi

christianschmucker:
Fuzzy search would be obviously better!
My implementation is quite rudimentary. ;)

TG908:
@nerdsupremacist for my refactor is basically started from scratch. I didn't get around to implementing search yet.

TUM Campus App/TUMSexy/TUMSexyTableViewController.swift
```diff
@@ -72,26 +81,52 @@ final class TUMSexyTableViewController: UITableViewController {
...
+        }
+        
         guard let urlString = link.target, let url = URL(string: urlString) else { return }
         UIApplication.shared.open(url)
     }
     
-}
+    // MARK: UISearchResultsUpdating
+    func updateSearchResults(for searchController: UISearchController) {
+        filteredData.removeAll(keepingCapacity: false)
```
TG908, line 122: Having two sources of truth is not that nice. I would suggest adding a `NSPredicate` to the `fetchedResultsController`. Something like this would be nice:

    // MARK: - UISearchResultsUpdating

    func updateSearchResults(for searchController: UISearchController) {
        guard let searchString = searchController.searchBar.text?.lowercased() else { return }
        let predicate = NSPredicate(format: "%K CONTAINS[cd] %@", #keyPath(TUMSexyLink.linkDescription), searchString)
        importer.fetchedResultsController.fetchRequest.predicate = predicate
        try? importer.fetchedResultsController.performFetch();
        self.tableView.reloadData()
    }

TG908 (changes requested):

PR #354, opened 2021-01-31
Parse tution fees with locale

### LINKED ISSUE #353: Student fees not parsed correctly
https://github.com/TUM-Dev/Campus-iOS/blob/5273980322a6a01b66998d3a8833112633ddd2c2/TUM%20Campus%20App/Tuition/Tuition.swift#L43

This seems to miss a german locale, the string "144,40" is parsed to 144. Originally reported at https://github.com/TUM-Dev/Campus-Android/issues/1380

Unsure if that's actually the underlying issue, never used swift.

joschahenningsen:
fixes #353

### Screenshots
before | after
---|---
Simulator Screen Shot - iPhone 11 - 2021-01-31 at 10 17 32 | Simulator Screen Shot - iPhone 11 - 2021-01-31 at 09 59 47

If we drop that error handling it would just display NaN in case of parsing errors. That might be fine as well because students could then at least see the deadline instead of just an error message.

### DIFF
--- a/TUM Campus App/Tuition/Tuition.swift
+++ b/TUM Campus App/Tuition/Tuition.swift
@@ -40,15 +40,16 @@ import CoreData
         let semester = try container.decode(String.self, forKey: .semester)
         let semesterID = try container.decode(String.self, forKey: .semesterID)
         let amountString = try container.decode(String.self, forKey: .amount)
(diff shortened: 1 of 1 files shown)


tgymnich (approved):
Thank you!

PR #369, opened 2021-09-28
created iso8601 calendar for determining weekOfYear

fkiwitt:
### Issue
If the user looks at the Mensa menu only the upcoming week will be displayed right now. This is caused by different calendar standards used in Campus iOS App and the Eat API resulting in a different weekOfYear than expected by the API. See https://en.wikipedia.org/wiki/Week#Other_week_numbering_systems for further information.

Calculating weekOfYear in the ISO8601 Calendar fixes this issues.


tgymnich (approved):
LGTM. Thanks a lot.

Issue #368, opened 2021-08-23
Switch Calendar implementation


tgymnich:
We should use this package for our calendar going forward:
https://github.com/kvyatkovskys/KVKCalendar

tgymnich:
As far as I know, nobody is currently working on this. Thanks for helping.

mvitanov:
#372 created to resolve this issue

PR #372, opened 2021-11-11
Feature/migrate to kvk calendar

mvitanov:
### Issue
Switch Calendar implementation #368

Migration from CalendarKit to KVKCalendar for its extended functionality.

### Problems encountered

1. Cannot keep the same text style as in CalendarKit since KVKCalendar does not accept NSAttributeString, but plain string
2. No animations when swotching days within same week
3. [App-related] I believe we should include a common spinner for the app to prevent user confusion while waiting for the APIs to respond.

### Screenshots
Simulator Screen Shot - iPod touch (7th generation) - 2021-11-11 at 12 05 58


mvitanov:
resolves #368

Issue #374, opened 2021-12-01
Homescreen-widget for Mensa menu 


NikolaiMadlener:
Widget for home screen that displays menu of the day at the Mensa of the user's choice (or the Mensa that is nearest to the user's current location)

tgymnich:
Great idea. Linking #85

NikolaiMadlener:
working on it, can you assign this one to me :)?

tgymnich:
Closed in #384

Issue #85, opened 2017-04-17
Today Extension / Widget


tgymnich:
I would like to create a Today Extension (Widget) for some of the cards like:

- News
- Grades
- Cafeteria
- Tution
- Calendar
-...

**Problems:**
1.  How should I name the new Targets (one for each card)
2. How should we share Code between both Targets
3. How should we share Data between both Targets
4. Dealing With CocoaPods will be fun

**Possible solutions:**
1. NewsWidget, GradesWidget, ...
2. Put our code in a Framework and share the Framework **OR** Add our source files to the app extension target (In both cases a few classes have to be rewritten/customized for the app extension)
3. Create an App Group to share data (name?)

tgymnich:
Closed in #384

PR #384, opened 2021-12-13
Nikolai/374 mensa menu widget

NikolaiMadlener:
### Issue

Widget for cafeteria plan (#374). Shows menu plan for current date and for selected cafeteria (long press the widget allows to select/set the cafeteria). Features dark & light mode. 

#374 #85

### Screenshots

Simulator Screen Shot - iPhone 13 Pro - 2021-12-13 at 14 10 05


TUM Campus App/Cafeteria/Dish.swift
```diff
@@ -54,7 +56,7 @@ struct Dish: Decodable {
      */
 
     let name: String
-    let prices: [String: Price]
+    let prices: [String: Price?]
```
tgymnich, line 59: This will prevent some Dishes from being parsed. See: https://github.com/TUM-Dev/eat-api/pull/26

TUM Campus App/Extensions/Extensions.swift
```diff
@@ -93,23 +93,6 @@ extension CodingUserInfoKey {
     static let context = CodingUserInfoKey(rawValue: "context")!
```
tgymnich, line 93: Thanks for cleaning this up 👍

MensaMenuWidget/MensaService.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import Foundation
+
+final class MensaService {
+    static let shared = MensaService()
+
+    private init() {}
+    
+    public func getMensaMenu(mensaApiKey: String) async throws -> Menu? {
+        let url = URL(string: "https://tum-dev.github.io/eat-api/" + "\(mensaApiKey)/\(Date().year)/\(Date().weekOfYear).json")!
```
tgymnich, line 17: Our API model can handle this already:
```suggestion
          let url = try EatAPI.menu(location: location, year: Date().year, week: Date().weekOfYear).asURLRequest().url
```

MensaMenuWidget/MensaService.swift
```diff
@@ -0,0 +1,36 @@
...
+        
+        let (data, _) = try await URLSession.shared.data(from: url)
```
tgymnich, line 19: Using Alamofire might be a bit nicer here.

```suggestion
         let url = EatAPI.menu(location: location, year: Date().year, week: Date().weekOfYear).asURLRequest()
  
          let decoder = JSONDecoder()
          let formatter = DateFormatter.yyyyMMdd
          decoder.dateDecodingStrategy = .formatted(formatter)
  
          AF.request(url).responseDecodable(of: MealPlan.self, decoder: decoder) { response in
              guard let mealPlan = response.value else { return }
              // ...
    }
```

tgymnich (changes requested):
Looks very nice. I left some small suggestions. 👍

TUM Campus App/Cafeteria/Dish.swift
NikolaiMadlener: Got it. The problem is that when this is not optional, my JSON decoder fails for some reason. not sure why...

NikolaiMadlener: Why will this prevent dishes from being parsed? The link you posted shows that some dished have prices that look like 
"prices": {
                            "students": null,
                            "staff": null,
                            "guests": null
                        }
which means in the code above Price should be optional right?
Also when I remove the optional, my JSON decoder can no longer decode it...
So I see no solution for this besides leaving it optional, maybe you can help me here?

tgymnich: My bad. I thought it was the other way around. This change should already be present in main. So a rebase / merge might a good idea before merging this.

tgymnich (approved):
Thanks!

PR #389, opened 2022-01-12
Fix broken mensa tab

Philipp000:
### Issue
The mensa tab is currently not working due to two issues:
- eat-api format has changed to use labels instead of ingredients on dish level
- For weeks 1-9 a leading `0` is required, to fetch the data

This PR fixes the two issues, and also loads the `labels.json` from the API, in order to show the short version of the (ingredient) labels

This fixes the following issue(s): 385

### Screenshots
Dish view:
image

Widget:
image

### TODO
I will include a popup, in order to show the explanations for the (ingredient) labels, similar to the popup on the website.

Fixes #385

PR #409, opened 2022-02-09
Add labels to the mensa dishes

Philipp000:
### Issue
Show the labels of the dishes in the cafeteria view and explain them.
Instead of a custom view, I chose to add the label descriptions to the disclosure group of a dish.

This PR is related to the labels introduced in #389 but is now based on the swiftUi branch.

### Screenshots
How it looks:
Bildschirmfoto 2022-02-09 um 14 08 04

Actually, I tried to have the label explanations and extended prices side-by-side, but this broke, when the price format is too long:
Bildschirmfoto 2022-02-09 um 14 11 44
If there are other ideas for the presentation, I would like to hear them.


tgymnich (approved):

Philipp000:
I have now replaced the icon and removed the prices for other groups than students

image

PR #413, opened 2022-02-14
Mv design calendar events

mvitanov:
### Issue
Calendar View migrated to SwiftUI using KVKCalendar

Features/Issues: 

- Different Calendar Types (day, week, month)
- CalendarEvents showed on all calendars
- Today Button functionality on all calendars
- List button using calendar list of events
- CalendarSingleEvent navigation view when an event is clicked from the calendar or the list

- Removes list when cancel is pressed on search screens
- Tuition open amount is now showed in red in the profile view section at the place of ✅
- Adding actionSheet to choose a speaker in LectureView and forwarding to lecture search with searchedText the chosen speaker from the sheet


AW-tum:
looks good and works fine on my simulator
didn't you want to remove the option for "week"?

mvitanov:
I kinda liked it how it looks to have a week overview of the events by hours -> you cannot have that neither in day mode, nor in month mode. The only thing I don't like is that you are still able to select days in week mode (but it does not really do anything).

AW-tum:
Hmm I get that
But in weekly-view I can't even read the courses titles or the times because it's all really squashed together (I attached an image below)
Maybe we could think about showing that option in landscape mode instead?
Like when you select days and then turn your phone it shows the whole week
But only then
Otherwise landscape mode is turned off
Because then it could be pretty useful and look pretty cool

Bildschirmfoto 2022-02-14 um 15 29 30

mvitanov:
This would certainly make it more readable, but allowing landscape mode only on the day calendar is quite non-intuitive. Me as a user would not know such an option exists. The idea of the week calendar is not to read the events, but rather to see an overview of the current week and which days are more intense, then you can go to the day calendar and choose the days, or simply press on the blue boxes -> they also lead to the event details page.

AW-tum:
fair enough
could we enable landscape mode automatically as soon as the user clicks on "week"?

mvitanov:
We will think about it. I think a good idea would be to group the events by colors and show all events of same courses in same color. Additionally for the week mode, we might show only the first letter of an event (or 3 letters or no titles at all). I guess this would provide a nice overview of a student's week schedule.

PR #416, opened 2022-03-08
SwiftUi redesign menu view in map

Philipp000:
### Issue
I think the implementation on how to access a menu in the swift-ui map is currently not very intuitive, with the small toolbar on the top.

Here I implemented a behavior which is more similar to the one used in Apple "Where is?" app.
Therefore, I completely remove the toolbar.
Further actions could be implemented directly into the panel.

Additionally, I fixed a few issues regarding the dark mode and the menu view.

### Screenshot
https://user-images.githubusercontent.com/1690395/157281029-01a94ab4-0375-415d-8312-d91445c760a4.mov


philippzagar:
@Philipp000 What's the current status of the PR? Can we merge it soon? :)

Philipp000:
From my side it should be ready to merge. I think @AW-tum wanted to merge it/look it at

Campus-iOS/Map/Cafeterias/MealPlanViewModel.swift
```diff
@@ -21,6 +21,8 @@ final class MealPlanViewModel: ObservableObject {
     init(cafeteria: Cafeteria) {
         self.cafeteria = cafeteria
         self.title = cafeteria.name
+        
+        fetch()
```
philippzagar, line 25: I would suggest calling `fetch()` not upon the initialization of the ViewModel but something like the `.task {}` modifier :)

Campus-iOS/Map/Cafeterias/MealPlanViewModel.swift
```diff
@@ -21,6 +21,8 @@ final class MealPlanViewModel: ObservableObject {
...
     }
     
     func fetch() {
```
philippzagar, line 28: Also make sure of some error handling here if this fetch request can't be done (eg. no network connection) -> Display some kind of `ErrorView` like can be seen in the `GradesScreen` etc.

Campus-iOS/Map/MainView/MapContent.swift
```diff
@@ -25,8 +25,10 @@ struct MapContent: UIViewRepresentable {
     @Binding var zoomOnUser: Bool
```
philippzagar, line 25: Do we really need this `UIViewRepresentable` here? Shouldn't we move away from UIKit?

Campus-iOS/Map/MainView/PanelContent.swift
```diff
@@ -150,8 +177,6 @@ struct SearchBar: View {
         PanelContent(zoomOnUser: .constant(true),
```
philippzagar, line 177: Why is the Preview actually commented out?

Campus-iOS/Map/MainView/PanelContent.swift
```diff
@@ -49,56 +88,46 @@ struct PanelContent: View {
...
                 }
-                .onChange(of: selectedCanteenName) { newValue in
-                    print("SELECTED CANTEEN (Panelcontent): ", newValue)
+                .searchable(text: $searchString, prompt: "Look for something")
+                .listStyle(PlainListStyle())
+            }
+        }
+        .onChange(of: selectedCanteen) { optionalCafeteria in
+            if let cafeteria = optionalCafeteria {
+                mealPlanViewModel = MealPlanViewModel(cafeteria: cafeteria)
```
philippzagar, line 112: As mentioned before, I guess there are cleaner ways to do that. Always instantiating a completely new ViewModel and then making the request upon the creation is a little overkill

Campus-iOS/Map/MainView/Panel.swift
```diff
@@ -94,9 +96,9 @@ enum PanelPosition: CGFloat {
     case top, middle, bottom
     var rawValue: CGFloat {
         switch self {
-        case .top: return 0.5 * screenHeight/3
-        case .middle: return 1.5 * screenHeight/3
-        case .bottom: return 3.2 * screenHeight/4
+        case .top: return (1/6) * screenHeight
```
philippzagar, line 99: Did u also test these Positions on an iPhone SE (4") / iPhone 13 Pro Max? (so the extreme screen sizes)

Campus-iOS/Map/MainView/Panel.swift
```diff
@@ -26,12 +24,16 @@ struct Panel: View {
...
-            PanelContent(zoomOnUser: $zoomOnUser,
+            VStack{
+                PanelContent(zoomOnUser: $zoomOnUser,
                          panelPosition: $panelPosition,
                          canteens: $canteens,
-                         selectedCanteenName: $selectedCanteenName,
-                         selectedAnnotationIndex: $selectedAnnotationIndex,
                          selectedCanteen: $selectedCanteen)
+                
+                //TODO: check for better way, to make last items in list available
```
philippzagar, line 33: Left todo?

Campus-iOS/Map/MainView/MapViewModel.swift
```diff
@@ -0,0 +1,38 @@
...
+    @Published var zoomOnUser: Bool = true
+    @Published var panelPosition: String = "pushMid"
+    @Published var selectedCanteenName: String = " "
+    @Published var selectedAnnotationIndex: Int = 0
+    
+    private let endpoint = EatAPI.canteens
+    private let sessionManager = Session.defaultSession
+    
+    func fetchCanteens() {
+        sessionManager.request(endpoint).responseDecodable(of: [Cafeteria].self, decoder: JSONDecoder()) { [self] response in
```
philippzagar, line 24: Maybe also create a separate `Service` for this network request, as can be seen in the Grade screens

Campus-iOS/Map/MainView/MapView.swift
```diff
@@ -9,32 +9,26 @@ import SwiftUI
...
+            Panel(zoomOnUser: $viewModel.zoomOnUser,
+                  panelPosition: $viewModel.panelPosition,
+                  canteens: $viewModel.canteens,
+                  selectedCanteen: $viewModel.selectedCanteen)
         }
-        .edgesIgnoringSafeArea(.all)
+        .edgesIgnoringSafeArea(.vertical)
         .navigationTitle("Map")
         .navigationBarHidden(true)
+        .onAppear() {
```
philippzagar, line 29: I guess the `task` modifier would be more state of the art. Also, think about the `async-await` functionality as this is a synchronous call :) I guess this is why u needed to capture `self` in the callback of the `MapViewModel `.

Campus-iOS/Map/MainView/MapContent.swift
```diff
@@ -91,21 +98,20 @@ struct MapContent: UIViewRepresentable {
...
                         latitude: locValue.latitude, longitude: locValue.longitude)
                     let span = MKCoordinateSpan(latitudeDelta: 0.01, longitudeDelta: 0.01)
                     let region = MKCoordinateRegion(center: coordinate, span: span)
                     
-                    mapView.setRegion(region, animated: true)
-                    for i in mapView.annotations {
-                        if i.title == selectedCanteenName {
-                            mapView.selectAnnotation(i, animated: true)
-                        }
+                    DispatchQueue.main.async {
```
philippzagar, line 111: Is there a better way to do that?

Campus-iOS/Map/MainView/MapContent.swift
```diff
@@ -53,6 +52,14 @@ struct MapContent: UIViewRepresentable {
         focusOnUser(mapView: view)
         focusOnCanteen(mapView: view)
         
+        if canteens.count > 0 && setAnnotations {
+            DispatchQueue.main.async {
```
philippzagar, line 56: Again, is there a better way to do that?

philippzagar (changes requested):
Really nice PR, thanks! :)
Design-wise it looks basically perfect! I only left some comments regarding the concrete implementation of it, mostly some state-of-the-art techniques.

philippzagar:
Also, a thing I'm just thinking of now: Maybe we should cut down the ingredients displayed, as it may be confusing and not really comprehensible. @AW-tum what's your take on this?

Campus-iOS/Map/MainView/MapContent.swift
Philipp000: I tried to use the SwiftUI map, however there were some difficulties with the showing of the pins/markers.

While it is possible to show these, they are not grouped together when they are close for example.
Also, the labels of markersmay overlap.

While it may be possible, to fix these, I felt that it would take very long, for a feature thats currently already implemented.

Campus-iOS/Map/MainView/PanelContent.swift
Philipp000: Actually, all the previews here are commented out, but I don't know what they are used for nor who created/commented them out.

Campus-iOS/Map/MainView/Panel.swift
Philipp000: There was a bug, that the last item in the panel was not properly shown. 
I fixed this, by adding empty space to the panel (which is then hidden behind the navigation bar).

I tried for several hours, to change the swiping behavior of the panel to be correct, but didn't manage to get it better.
So the current version kind of works, but is not ideal. However, I don't know how to make it better.

Campus-iOS/Map/MainView/PanelContent.swift
Philipp000: What could be a cleaner way? 
I felt that reusing the view model and only updating it is even worse, but I didn't know how to do it in another way.

Campus-iOS/Map/MainView/MapContent.swift
Philipp000: Actually, I really don't know, but it was also implemented with the DispatchQueue before.
If you have a better idea, feel free to let me know.

Philipp000:
What do you mean about the ingredients?
There was already feedback, and the ingredients are only shown, when the info is clicked (not like in the video anymore).

I have also another small issue, that I wasn't able to fix: 
When one selects a canteen and then closes the menu again, the "scroll-position" is not remembered. So, the canteen list is shown from the top again. 
However, I have no idea on how to fix this.

Campus-iOS/Map/MainView/Panel.swift
Philipp000: The positions are the same as before, I just thought (1/6) * x is more readable than 0.5 * x / 3.
However, I haven't tested them again, as I thought they should be okay.

Campus-iOS/Map/MainView/PanelContent.swift
philippzagar: They are used for the SwiftUI previews, feel free to comment all of them in again :D

Campus-iOS/Map/MainView/PanelContent.swift
philippzagar: Well, just update the data in the viewmodel right? No need to allocate a completly new object.

Campus-iOS/Map/MainView/Panel.swift
philippzagar: If possible, please test that as it could have been wrong before :)

Campus-iOS/Map/MainView/MapContent.swift
philippzagar: Why is the async DispatchQueue actually required here? I'm not really getting it 😅

Campus-iOS/Map/MainView/MapContent.swift
philippzagar: K that seems like a reasonable tradeoff, lets leave it like that for now

AW-tum:
I just had a look at it in the simulator and I have a few UI remarks

First of all I would stretch the calendar view of the days along the whole screen (right now it seems a little squashed in the center). 

Also I would handle the ingredients a little differently. When you open the ingredient subview you get a list of the ingredients as well as their icons. The point of the icons is that we don't have to write out the ingredients every time and therefore save space onscreen.
I would put the icons inside the table cells and remove the possibility to open a subview for ingredients (or show only the icons inside the subview). Instead I would put an "I" button at the top of the card, next to the dismiss-button which gives you a description of all the ingredients and their corresponding icons.

Today (Sunday) obviously no menu can be displayed. For me at least, it hence portrays a blank screen below the calendar-bar. Instead of showing a blank screen I would either show some form of "no menu today"/"cafeteria closed on weekends"/etc. or show the next available menu (in todays case, the one of upcoming Monday). We could signal to the user that it's not actually todays menu by highlighting the menus date with the black circle, setting it apart from a TUM-blue circle to highlight the usage dates day.

philippzagar:
> What do you mean about the ingredients?
> 
> There was already feedback, and the ingredients are only shown, when the info is clicked (not like in the video anymore).
> 
> 
> 
> I have also another small issue, that I wasn't able to fix: 
> 
> When one selects a canteen and then closes the menu again, the "scroll-position" is not remembered. So, the canteen list is shown from the top again. 
> 
> However, I have no idea on how to fix this.

See @AW-tum remarks

PR #455, opened 2022-07-06
Improve Movies Page

Atharva-Mathapati:
### Issue
Movies page did not match other pages in terms of UI

Added features:
- match Lectures Page UI
- filtered movies list so that only upcoming movies are shown
- added Parallax Effect


AW-tum:
I like the design a lot 
Also what you did with the navigation bar appearing when scrolling down
I will add a few minor Design tweaks

AW-tum:
Also I just realized that the Basic Information bit that displays a clock does not show any information for any film
Is that sth to be added in the code or the API?
Otherwise I would remove this part

AW-tum:
Same goes for the location
For every movie it just shows "location" without any information no the actual location

AW-tum:
already a lot better
now it's also more intuitive that if you dismiss the sheet manually by pulling it down it dismisses back to the movies overview and not the Homescreen
now all that's needed is some form of dismiss button

PR #500, opened 2022-10-12
Fix/map panel

AW-tum:
Main fixes
- map panel now draggable all the way to the bottom
- increased drag area
- cafeteria annotations reappearing after actuating picker


14slash12 (approved):
LGTM! 😊

PR #522, opened 2023-01-28
Simple Crashlytics Service

Antonwy:
Instead of:
```swift
#if !DEBUG
            Crashlytics.crashlytics().record(error: error)
#endif
```

You can now write just:
```swift
CrashlyticsService.log(error)
```

### DIFF
--- a/Campus-iOS/Crashlytics/CrashlyticsService.swift
+++ b/Campus-iOS/Crashlytics/CrashlyticsService.swift
@@ -1,0 +1,25 @@
+//
+//  CrashlyticsService.swift
+//  Campus-iOS
+//
+//  Created by Anton Wyrowski on 28.01.23.
+//
+
+import Foundation
+import FirebaseCrashlytics
+
+class CrashlyticsService {
+    static private let crashlytics = Crashlytics.crashlytics()
+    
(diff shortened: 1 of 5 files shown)


philippzagar (approved):
LGTM! Thanks!

Campus-iOS/Crashlytics/CrashlyticsService.swift
```diff
@@ -0,0 +1,25 @@
...
+//  CrashlyticsService.swift
+//  Campus-iOS
+//
+//  Created by Anton Wyrowski on 28.01.23.
+//
+
+import Foundation
+import FirebaseCrashlytics
+
+class CrashlyticsService {
```
philippzagar, line 11: One minor thing: One could also put the call to the config method from Crashlytics into this service

Antonwy: I thought of this too, but the config call is `FirebaseApp.configure()`. 
So the call is not only for crashlytics, but also for all other services of firebase and therefore perhaps wrong in the crashlytics service.

philippzagar: Seems reasonable 👍

Issue #524, opened 2023-03-08
Add SiriKit shortcuts donations for suggestions


ge65cer:
- Donate shortcuts when user performs action in the app (e.g., looks up food for specific mensa)
- Add deep links to specific actions
- Used for example to suggest widgets in Smart Stack feature 
- Allows Siri to suggest common actions (e.g., at noon recommend mensa plan)

https://developer.apple.com/videos/play/wwdc2022/10032/

<img width="500" title="Example" src="https://media.idownloadblog.com/wp-content/uploads/2022/06/Siri-Suggestions-widget-iPhone.jpg" />

PR #526, opened 2023-03-09
Add app intents

ge65cer:
### Issue

https://developer.apple.com/documentation/appintents

This fixes the following issue(s): #524


ge65cer:
Closed due to missing reviews.

PR #538, opened 2023-04-04
Fix Swift 5.8 compile errors

ge65cer:
### Issue

Xcode 14.3 shipped with Swift 5.8, which resulted in breaking changes in our codebase.
`Initializer for conditional binding must have Optional type`

See also: https://github.com/apple/swift/issues/63808
More info: https://forums.swift.org/t/improved-result-builder-implementation-in-swift-5-8/63192


philippzagar (approved):
LGTM! Thanks for the quick fix, i already got notified yesterday by someone else regarding this issue :)

Issue #553, opened 2023-10-16
Empty state for grades view


mohanadkandil:
**Describe the bug**
There's an empty view for students without grades (1st semester)

**To Reproduce**
1. Go to Grades view on any platform
2. Empty white/black page

<p align="center"> 
  <img src="https://github.com/TUM-Dev/Campus-Flutter/assets/18647990/b8dfda1f-9562-40e3-b5ce-8ce381d8ac8a" alt="grades view state" width="200" />
</p>

**Expected behavior**
A clear SVG/Diagram like the one in Calendar state that "No grades until now"

<p align="center">
  <img src="https://github.com/TUM-Dev/Campus-Flutter/assets/18647990/376e0897-1eed-4795-bb9d-1e8c4fed4fd7" alt="calendar empty state" width="200" align="center" />
</p>

jakobkoerber:
Hey, you seem to be in the wrong repository :D therefore I transfered the issue to Campus-iOS.

Could you try to install the beta via [TestFlight](https://testflight.apple.com/join/4Ddi6f2f) and test if the issue still persists? If so, please create a fork for your bugfix. We can later merge that into the repo.

mohanadkandil:
@jakobkoerber thanks for letting me know. I am testing it now. Could you please send an invitation code for TestFlight ?

jakobkoerber:
The code is provided via the link

mohanadkandil:
Yeah it's an empty white/black background

<img src="https://github.com/TUM-Dev/Campus-iOS/assets/18647990/3e9afdc2-af8f-4542-8b0c-5caf3c8267ad" alt="ios-grades-view" width="300" />

jakobkoerber:
Interesting, are you logged in? If you want to you can create a fork with the fix. Thanks for letting us know!

mohanadkandil:
I'll start working on it. 
Can I use the same SVG for the calendar, custom one,  or what do you recommend for the grades view ?

jakobkoerber:
please use the same

jakobkoerber:
> **Expected behavior** A clear SVG/Diagram like the one in Calendar state that "No grades until now"
> 
> <img alt="calendar empty state" width="200" src="https://user-images.githubusercontent.com/18647990/275624329-376e0897-1eed-4795-bb9d-1e8c4fed4fd7.PNG">

[...]

PR #555, opened 2023-10-17
Fix empty state for grades view

mohanadkandil:
### Issue
Add an empty state for the grades view. 

Simple solution:
Just check that the grades is <= 0 and then show the Failed screen with "No grades yet!" (as an error message)

#553 

### Screenshots
grades empty state


jakobkoerber:
Thanks for contributing! Two requests from my side:
- could you maybe add this to all tab views? this would help to add some consistency
- I don't think that the FailedView is applicable here, it hasn't failed, the user just does not have any grades... maybe we should add a simple centered text saying something like "You seem to not have any grades yet!" or similar

jakobkoerber (changes requested):
Thanks for contributing! Two requests from my side:

- could you maybe add this to all tab views? this would help to add some consistency
- I don't think that the FailedView is applicable here, it hasn't failed, the user just does not have any grades... maybe we should add a simple centered text saying something like "You seem to not have any grades yet!" or similar

mohanadkandil:
I just added a text here but seems the top bar is collapsed 

Screenshot 2023-10-20 at 1 05 13 AM

mohanadkandil:
> * d help to add some con



> Thanks for contributing! Two requests from my side:
> 
> * could you maybe add this to all tab views? this would help to add some consistency
> * I don't think that the FailedView is applicable here, it hasn't failed, the user just does not have any grades... maybe we should add a simple centered text saying something like "You seem to not have any grades yet!" or similar

For the 1st point.
Do you mean to add the same logic of handling the grades (if it's <= 0) in all views ? 
if yes so what view like places (grades are not handled there afaik)

jakobkoerber:
> For the 1st point. Do you mean to add the same logic of handling the grades (if it's <= 0) in all views ? if yes so what view like places (grades are not handled there afaik)

No, I think that it would be also useful to handle this behaviour with lectures

mohanadkandil:
I just added a view for the empty grades (it's the best way to avoid collapsed menus), I might work on separate PR for refactoring it.

mohanadkandil:
> > For the 1st point. Do you mean to add the same logic of handling the grades (if it's <= 0) in all views ? if yes so what view like places (grades are not handled there afaik)
> 
> No, I think that it would be also useful to handle this behaviour with lectures

But the lecture view doesn't have anything related to grades, soi why we are trying to handle this case in it ?

jakobkoerber:
> But the lecture view doesn't have anything related to grades, soi why we are trying to handle this case in it ?

The lecture view can have no lectures, therefore we should handle that similarly to the grades view.

Campus-iOS/GradesComponent/Screen/GradesScreen.swift
```diff
@@ -26,63 +26,73 @@ struct GradesScreen: View {
...
-            switch vm.state {
-            case .success(_):
-                VStack {
-                    GradesView(
-                        vm: self.vm
-                    )
-                    .padding(.top, 30)
-                    .refreshable {
-                        await vm.reloadGradesAndAverageGrades(forcedRefresh: true)
+        if (vm.grades.count > 0) {
```
jakobkoerber, line 29: please remove that if else clause - applying this before data is fetched, makes no sense

Campus-iOS/GradesComponent/Screen/GradesScreen.swift
```diff
@@ -26,63 +26,73 @@ struct GradesScreen: View {
...
+                    if case let .failed(error) = detail {
+                        if let apiError = error as? TUMOnlineAPIError {
+                            Text(apiError.errorDescription ?? "TUMOnlineAPI Error")
+                        } else {
+                            Text(error.localizedDescription)
+                        }
                     }
                 }
-            }
+        } else {
```
jakobkoerber, line 92: please remove that if else clause - applying this before data is fetched, makes no sense

Campus-iOS/Base/Views/EmptyGradesView.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+
+
+import SwiftUI
+
+struct EmptyGradesView: View {
+    var body: some View {
+        GeometryReader { geo in
+            ZStack {
+                Image(decorative: "Error-logo-transparent")
```
jakobkoerber, line 15: It's questionable for me if the error image is applicable here, since it's not an error

Campus-iOS/Base/Views/EmptyGradesView.swift
```diff
@@ -0,0 +1,33 @@
...
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .offset(y: -geo.size.height*0.1)
+                
+                VStack (alignment: .center, spacing: 12) {
+                    Text("You seem to not have any grades yet!")
```
jakobkoerber, line 21: please localise this string

Campus-iOS/Base/Views/EmptyGradesView.swift
```diff
@@ -0,0 +1,33 @@
...
+//  TestView.swift
+//  Campus-iOS
+//
+//  Created by Mohanned Kandil on 23.10.2023.
+//
+
+
+import SwiftUI
+
+struct EmptyGradesView: View {
```
jakobkoerber, line 11: we should make it reusable for empty lectures

Campus-iOS/GradesComponent/Screen/GradesScreen.swift
mohanadkandil: it's already in the `success` case, cause the incoming data in the empty case is `[]`, which is considered ti the success case. That's whyI handled the main logic in success

jakobkoerber: please remove it at line 29 and 92

jakobkoerber:
the behaviour now is:
|Empty Lectures|Empty Grades|
|-|-|
Simulator Screenshot - iPhone 15 Pro - 2023-10-23 at 17 56 52 | Simulator Screenshot - iPhone 15 Pro - 2023-10-23 at 17 56 48

jakobkoerber (approved):

